PR #5, opened 2024-08-30
[Jonas Seet] iP

Dri-water:


src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
+import java.util.Scanner;
+
+public class Archibald {
+
+    private static Task[] tasks = new Task[100];
+    private static int taskCount = 0;
```
YukeeHong: instead of taskCount, consider using totalTasks for better clarity.

src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
...
+                unmarkTask(input);
+            } else {
+                addTask(input);
+            }
+        }
+
+        scanner.close();
+    }
+
+    public static void markTask(String input) {
```
YukeeHong: Unmark ang mark is similar. Refactor shared code into a helper method to reduce duplication.

src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
...
+            return;
+        }
+
+        String[] parts = input.split(" ", 2);
+        String type = parts[0].toLowerCase();
+        String details = parts.length > 1 ? parts[1] : "";
+
+        switch (type) {
+            case "todo":
+                tasks[taskCount++] = new Todo(details);
```
YukeeHong: There should be consistent spaces around operators to improve readability.

src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
...
+                break;
+            case "deadline":
+                String[] deadlineParts = details.split(" /by ", 2);
+                tasks[taskCount++] = new Deadline(deadlineParts[0], deadlineParts[1]);
+                break;
+            case "event":
+                String[] eventParts = details.split(" /from | /to ");
```
YukeeHong: The split logic assumes that the /from and /to keywords will always be formatted correctly. If the user doesn't follow this format exactly, it may lead to an ArrayIndexOutOfBoundsException.

src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
+import java.util.Scanner;
+
+public class Archibald {
+
+    private static Task[] tasks = new Task[100];
```
ThisisXXZ: Consider defining a constant for the array size. For example, ``private static final int MAX_TASKS = 100``

src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
...
+        System.out.println("____________________________________________________________");
+    }
+
+    public static void addTask(String input) {
+        if (taskCount >= tasks.length) {
+            printArchibaldResponse("Error: Task list is full!");
+            return;
+        }
+
+        String[] parts = input.split(" ", 2);
```
ThisisXXZ: When splitting the input and accessing parts[0], there is a risk of ``ArrayIndexOutOfBoundsException``. Maybe could add some validations before accessing the array?

src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
...
+        String type = parts[0].toLowerCase();
+        String details = parts.length > 1 ? parts[1] : "";
```
ThisisXXZ: I like how you use the ternary operator here!

src/main/java/Archibald.java
```diff
@@ -0,0 +1,104 @@
...
+
+    public static void printList() {
+        printArchibaldResponse("Here are the tasks in your list:");
+        for (int i = 0; i < taskCount; i++) {
+            System.out.println((i + 1) + ". " + tasks[i]);
+        }
+    }
+
+    public static void main(String[] args) {
+        String name = "Archibald";
```
ThisisXXZ: Since the name of the chatbox remains unchanged, it may be more appropriate to define the ``name`` as a constant ``NAME``

ThisisXXZ (commented):
ThisisXXZ: PR Review for Jonas Seet

src/main/java/archibald/Archibald.java
```diff
@@ -0,0 +1,206 @@
...
+                    break;
+                case "event":
+                    String[] eventParts = parts[1].split(" /from | /to ");
+                    if (eventParts.length < 3) {
+                        throw new ArchibaldException("Error: Event must include '/from' and '/to' followed by dates.");
+                    }
+                    newTask = new Event(eventParts[0], eventParts[1], eventParts[2]);
+                    break;
+                default:
+                    throw new ArchibaldException("Error: In spite of my knowledge... I don't know what that means >:(");
```
wenxin-c: Based on requirements, there shouldn't be indentation in front of case clause. Please take note for all the switch below as well.
E.g.
```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}
```

src/main/java/archibald/Archibald.java
```diff
@@ -0,0 +1,206 @@
...
+            }
+    
+            tasks.add(newTask);
+            printArchibaldResponse("Got it. I've added this task:\n  " + newTask +
+                                   "\nNow you have " + tasks.size() + " tasks in the list.");
```
wenxin-c: Indentation before wrapped lines should be 8 spaces.
E.g. 
```
setText("Long line split"
        + "into two parts.");
if (isReady) {
    setText("Long line split"
            + "into two parts.");
}
```

src/main/java/archibald/Archibald.java
```diff
@@ -0,0 +1,206 @@
...
+                }
+            }
+    
+            Task newTask;
+            switch (type) {
+                case "todo":
+                    newTask = new Todo(parts[1]);
+                    break;
+                case "deadline":
+                    String[] deadlineParts = parts[1].split(" /by ", 2);
```
wenxin-c: Try to avoid magic literals, you can consider using named constants to give the literals some meaning so that it will be easier to understand as well. 
E.g.
```
static final double PI = 3.14236;
static final int MAX_SIZE = 10;
...
return PI;
...
return MAX_SIZE - 1;

```

src/main/java/archibald/Archibald.java
```diff
@@ -0,0 +1,206 @@
...
+            printArchibaldResponse("Error: Unable to load tasks from file.");
+        }
+    }
+
+    public static void deleteTask(String input) {
+        try {
+            int taskIndex = Integer.parseInt(input.split(" ")[1]) - 1;
+            if (taskIndex >= 0 && taskIndex < tasks.size()) {
+                Task removedTask = tasks.remove(taskIndex);
+                printArchibaldResponse("Duly noted sire. Thy hath removed this task:\n  " + removedTask + "\nNow thou hath " + tasks.size() + " tasks in the list.");
```
wenxin-c: Consider wrapping around the sentence if it gets too long

src/main/java/archibald/Archibald.java
```diff
@@ -0,0 +1,206 @@
+package archibald;
+import java.util.Scanner;
+import java.io.*;
```
wenxin-c: You should import the libraries individually instead of using a wildcard (i.e. *).

wenxin-c (commented):
Overall good job so far! Just be more mindful of the coding standard and quality requirements. You can find them on course website. [Coding Standard](https://se-education.org/guides/conventions/java/basic.html), [Code quality](https://nus-cs2113-ay2425s1.github.io/website/se-book-adapted/chapters/codeQuality.html)

PR #7, opened 2024-08-30
[Ethan Chua] iP

RCPilot1604:


src/main/java/Erika.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+    }
+
+    private static void addEvent(String line) {
+        int indexOfFrom = line.indexOf("/from ");
+        int indexOfTo = line.indexOf("/to ");
+        if (indexOfFrom != -1 && indexOfTo != -1) {
+            Event newEvent = new Event(line.substring(line.indexOf(" ")+1, indexOfFrom-1),
+                    line.substring(indexOfFrom+6, indexOfTo-1), line.substring(indexOfTo+4));
```
wahkia: spacing issues

few more similar issues below

src/main/java/Erika.java
```diff
@@ -0,0 +1,162 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Erika {
+    private static ArrayList<Task> tasks = new ArrayList<>();
+    private static Scanner in = new Scanner(System.in);
+    private static int markIndex;
```
wahkia: Variable declared follow the camelCase convention, adhering to the Java naming standards. This is good!

src/main/java/Erika.java
```diff
@@ -0,0 +1,162 @@
...
+                addDeadline(line);
+            } else if (line.contains("event ")) {
+                addEvent(line);
+            } else {
+                printUnknownCommand();
+            }
+        }
+    }
+
+    private static void addEvent(String line) {
```
wahkia: Maybe you could include some Javadoc comments for the functions you created. It would enhance its maintainability and readability of the code.

src/main/java/Erika.java
```diff
@@ -0,0 +1,162 @@
...
+    private static int extractTaskIndex(String line) {
+        String digitString = line.replaceAll("[^0-9-]", "");
+        if (!digitString.isEmpty()) {
+            return Integer.parseInt(digitString);
+        } else {
+            return -1;
+        }
+    }
+
+    private static void printMessage(String message) {
```
wahkia: Nice function for printing!

wahkia (commented):
Overall, this is a well-written and structured piece of code that adheres to most coding standards!

src/main/java/Erika.java
RCPilot1604: Thanks for pointing this out!

src/main/java/Task.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public void finalize() {
+        taskArraySize--;
+    }
+
+    public String printString() {
+        return "";
+    }
```
AndrewNguyen4: - Class java.lang.Object has a default toString() method. You should override this method instead of creating a new method. This applies for all instances of printString() in this program.
- Confusing method name: I would expect printString() to print the string using System.out.println(), not return a string type object.

src/main/java/Erika.java
```diff
@@ -0,0 +1,162 @@
...
+            } else if (line.equals("list")) {
+                printList();
+            } else if (line.contains("mark ")) {
+                handleMark(line);
+            } else if (line.contains("todo ")) {
+                addTodo(line);
+            } else if (line.contains("deadline ")) {
+                addDeadline(line);
+            } else if (line.contains("event ")) {
+                addEvent(line);
```
AndrewNguyen4: Polymorphism could be used here to reduce the need for handling different cases.

src/main/java/Erika.java
AndrewNguyen4: The specific implementation of this method (and other methods specific to their task type, e.g. addDeadline(), addTodo()) should be put in the corresponding class to improve the effect of OOP in your code.

src/main/java/Erika.java
AndrewNguyen4: The markIndex member of this class can also serve to keep track of the number of tasks in the array, replacing taskArraySize member of classes below. This helps reduce the redundancy in using members.

RCPilot1604: Hi thanks for your comments! I would like to clarify this comment as markIndex is currently used to store the item the user wishes to mark via the "mark"/"unmark" command. I'm not sure how it can be recycled to also store the size of the task array. Thanks for any clarification!

src/main/java/Erika.java
RCPilot1604: Sorry I don't quite understand how said polymorphism will work since I need to deliberately create a new object of the respective subclass (Todo, Deadline, Event) when adding to the task array. Therefore I am not sure how I can leverage polymorphism to call a single (overriden) method for different cases. Thanks!

src/main/java/erika/task/Todo.java
```diff
@@ -0,0 +1,17 @@
+package erika.task;
+
+public class Todo extends Task{
+    public Todo (String description) {
+        super(description);
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[T][%s] %s", isDone ? "X":" ", description);
```
Mahesh1772, line 18: Consider using the parent class's `toString` method to apply inheritance principle

src/main/java/erika/task/Task.java
```diff
@@ -0,0 +1,54 @@
...
+    public void setDescription(String description) {
+        this.description = description;
+    }
+
+    public static int getTaskArraySize() {
+        return taskArraySize;
+    }
+
+    public static void setTaskArraySize(int size) {
+        taskArraySize = size;
```
Mahesh1772: Is a setter needed for this use case, since you already seem to have a increment and decrement method

src/main/java/erika/task/Event.java
```diff
@@ -0,0 +1,22 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[E][%s] %s (from: %s to: %s)", isDone ? "X":" ", description, start, end);
+    }
+
+    @Override
+    public String generateFileLine() {
+        return String.format("E,%s,%s,%s,%s\n", isDone ? "1" : "0", description, start, end);
```
Mahesh1772: Consider implementing the `generateFileLine()` method in the `Task` class and use inheritance for better scalability

src/main/java/erika/settings/Settings.java
```diff
@@ -0,0 +1,15 @@
...
+    public static final int BY_LENGTH_OFFSET = 4;
+    public static final int SPACE_OFFSET = 1;
+    public static final int BY_REAR_OFFSET = 1;
+    public static final int FROM_REAR_OFFSET = 1;
+    public static final int TO_REAR_OFFSET = 1;
+
+    public static final String DIRECTORY = "data";
+    public static final String FILENAME = DIRECTORY + "/tasks.txt";
+    public static final String SEPARATOR = ",";
+}
```
Mahesh1772: Good job defining constants to a file

src/main/java/erika/filesystem/FileSystem.java
```diff
@@ -0,0 +1,106 @@
...
+            Deadline deadline = new Deadline(description, tokens[3]);
+            deadline.setMark(Boolean.parseBoolean(tokens[1]));
+            tasks.add(deadline);
+            break;
+        case "E":
+            if (tokens.length != 5) {
+                throw new FileFormatErrorException("Error on line " + lineNumber + ": invalid Event line!");
+            }
+            Event event = new Event(description, tokens[3], tokens[4]);
+            event.setMark(Boolean.parseBoolean(tokens[1]));
```
Mahesh1772, line 173: Consider having a `setMark` method in the `Task` class as to reduce duplicate lines of code in different subclasses

src/main/java/erika/Erika.java
```diff
@@ -0,0 +1,293 @@
...
+        String digitString = line.replaceAll("[^0-9-]", "");
+        if (!digitString.isEmpty()) {
+            return Integer.parseInt(digitString);
+        } else {
+            throw new NumberFormatException();
+        }
+    }
+
+    private static void printMessage(String message) {
+        System.out.println("\t____________________________________________________________");
```
Mahesh1772: Consider making this line into a constant string, for easy reproducibility and keeping visuals consistent

src/main/java/erika/Erika.java
```diff
@@ -0,0 +1,293 @@
...
+    private static void addTodo(String line) throws EmptyDescriptionException{
+        if (!line.contains("todo ")) {
+            throw new EmptyDescriptionException("Todo");
+        }
+        Todo newTodo = new Todo(line.substring(line.indexOf(" ")+Settings.SPACE_OFFSET));
+        tasks.add(newTodo);
+        try {
+            fileSystem.appendTaskToFile(newTodo);
+        } catch (IOException e) {
+            printMessage("Unable to append Todo to FS. IO Error occurred");
```
Mahesh1772: Consider adding these error messages into the constants file as well

src/main/java/erika/Erika.java
```diff
@@ -0,0 +1,293 @@
...
+                addTodo(line);
+            } else if (line.contains("deadline")) {
+                addDeadline(line);
+            } else if (line.contains("event")) {
+                addEvent(line);
+            } else if (line.contains("delete")) {
+                indexOperation(line, true);
+            } else {
+                throw new UnknownCommandException();
+            }
```
Mahesh1772: Consider using switch case for clearer code organization and because the values expected are predetermined.

Mahesh1772 (commented):
Overall, this implementation is solid, with a clear understanding of task management logic, but there are areas for improvement. Consider adding **JavaDocs** in future commits. Keep up the good work!

PR #8, opened 2024-08-30
[DanLinhHuynh-Niwashi] iP

DanLinhHuynh-Niwashi:


src/main/java/Niwa.java
```diff
@@ -0,0 +1,322 @@
...
+        processCommand(COMMAND_HELP_WORD);
+    }
+
+    /**
+     * Processes the command entered by the user.
+     *
+     * @param command The user command to process.
+     */
+    public void processCommand(String command) {
+        System.out.println(SEPARATOR);
```
rharwo: Good use of final variables

src/main/java/Niwa.java
```diff
@@ -0,0 +1,322 @@
...
+            case COMMAND_DELETE_WORD:
+                delete(commandParts[1]);
+                break;
+            case COMMAND_EVENT_WORD:
+                // Fall through
+            case COMMAND_DEADLINE_WORD:
+                // Fall through
+            case COMMAND_TODO_WORD:
+                add(commandParts[1], commandParts[0]);
+                break;
```
rharwo: Good use of comments to make it clear that it falls through

src/main/java/Niwa.java
```diff
@@ -0,0 +1,322 @@
...
+
+    /**
+     * Lists all tasks in the task list.
+     */
+    private void list() {
+        System.out.println(PREFIX + COMMAND_LIST_SUCCESS);
+        int index = 1;
+
+        for (Task task : tasks) {
+            System.out.printf(PREFIX + "%d. %s%n", index++, task.getFullInfo());
```
rharwo: Unsure about this, but is it good to have index++ inside of the print statement? May make it hard to find if it causes an error.

src/main/java/Niwa.java
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    /**
+     * Adds a new task to the task list.
+     *
+     * @param taskInfo The task details to add.
+     * @param type The type of task to add ("event", "deadline", "todo").
+     */
+    private void add(String taskInfo, String type) {
+        Task temp;
```
rharwo: Is it good to use temp instead of something like toAdd?

rharwo (commented):
LGTM.

dZhei98 (commented):
No major issues found. Good job!

PR #11, opened 2024-08-30
[Si Thu Lin Aung] iP

C2linaung:


src/main/java/Functions.java
```diff
@@ -0,0 +1,142 @@
...
+    protected static final Scanner in = new Scanner(System.in);
+    private static final String SEPARATOR = "_________________________________________________________";
+    private Task[] taskList = new Task[100];
+    private int taskCounter = 0;
+
+    public Functions() {
+        printFunctions();
+    }
+
+    public void printFunctions(){
```
yeekian: I like that you abstracted out the messages so that the code is less wordy.

src/main/java/Functions.java
```diff
@@ -0,0 +1,142 @@
...
+            String input = in.nextLine();
+            processCommand(input);
+        }
+    }
+
+    private void processCommand(String input) {
+        if (input.startsWith("mark ") || input.startsWith("unmark ")) {
+            handleMarking(input);
+        }
+        else {
```
yeekian: Should "switch" and "case" in the switch case be indented by the same amount as per the coding standard?

src/main/java/Functions.java
```diff
@@ -0,0 +1,142 @@
...
+            taskList[taskCounter++] = new Deadlines(input);
+        }
+        else if (input.startsWith("event ")){
+            taskList[taskCounter++] = new Events(input);
+        }
+        else{
+            taskList[taskCounter++] = new Task(input);
+        }
+
+        System.out.println("Got it. I've added this task:");
```
yeekian: Should the printed messages be extracted out as a separate function for readability?

src/main/java/ToDos.java
```diff
@@ -0,0 +1,11 @@
+public class ToDos extends Task{
```
yeekian: I like your use of inheritance to build additional classes on top of Task class

src/main/java/Functions.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    private void processCommand(String input) {
+        if (input.startsWith("mark ") || input.startsWith("unmark ")) {
+            handleMarking(input);
+        }
+        else {
+            switch (input) {
+                case "bye":
```
yeekian: Should the "bye" case be put at the back so that other cases can seen first?

src/main/java/Functions.java
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+    }
+
+    private void handleMarking(String input) {
+        try {
+            int taskNumIndex = Integer.parseInt(input.substring(input.indexOf(' ') + 1)) - 1;
+            if (taskNumIndex >= 0 && taskNumIndex < taskCounter) {
+                if (input.startsWith("mark ")) {
+                    taskList[taskNumIndex].setDone();
```
yeekian: Should the print messages be extracted out further to ensure that the same level of abstraction is used?

yeekian (commented):
Good job on your code!

src/main/java/Events.java
```diff
@@ -0,0 +1,35 @@
+public class Events extends Task{
+
+    protected String from;
+    protected String to;
```
yakultbottle: Should this be made final?

src/main/java/Functions.java
```diff
@@ -0,0 +1,142 @@
...
+            System.out.println("Say anything! If you are no longer bored, type exit !");
+            String input = in.nextLine();
+            if(input.equals("exit")){
+                System.out.println(SEPARATOR);
+                printFunctions();
+                break;
+            }
+            System.out.println(SEPARATOR);
+            System.out.println(input);
+            System.out.println(SEPARATOR);
```
yakultbottle: Should System.out.println(SEPARATOR) be its own method, since you have it declared as private final in two different classes?

src/main/java/Main.java
```diff
@@ -0,0 +1,5 @@
+public class Main {
+    public static void main(String[] args) {
+        Chatbot uranus = new Chatbot();
```
yakultbottle, line 5: I like that the Chatbot functionality is abstracted into its own class

src/main/java/Events.java
```diff
@@ -0,0 +1,35 @@
...
+        this.from = from;
+    }
+
+    public String getTo() {
+        return to;
+    }
+
+    public void setTo(String to) {
+        this.to = to;
+    }
```
yakultbottle: Should these methods be collapsed into the constructor, since they are not used outside of it?

src/main/java/Task.java
```diff
@@ -0,0 +1,42 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    private String tag;
```
yakultbottle: Should tag be renamed as icon?

src/main/java/Task.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public Task(String description) {
+        this.description = description;
+        this.isDone = false;
+        this.tag = "N"; // for no-tag
```
tsoudibi: isn't no-tag should be blank instead of N?

src/main/java/Events.java
```diff
@@ -0,0 +1,35 @@
...
+
+    @Override
+    public void setDescription(String description){
+        int separatorIndex = description.indexOf(' ');
+        String[] str = description.split("/");
+        setFrom(str[1].substring(4));
+        setTo(str[2].substring(2));
```
tsoudibi: avoid magic numbers

src/main/java/Deadlines.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public String getBy(){
+        return by;
+    }
+
+    @Override
+    public void setDescription(String description){
+        int separatorIndex = description.indexOf(' ');
+        String[] str = description.split("/by");
```
tsoudibi: shouldn't name the variable "str"

tsoudibi: pick one between "" and ''

src/main/java/Chatbot.java
```diff
@@ -0,0 +1,33 @@
...
+    // Constructor for chatbot
+    public Chatbot() {
+        start();
+    }
+
+    // Starting Chatbot
+    public void start() {
+        printWelcomeMessage();
+        Functions execute = new Functions();
+        execute.taskmaster();
```
tsoudibi: LGTM, nice to use taskmaster

tsoudibi (commented):
looks pretty good, still small things to fix

src/main/java/Functions.java
C2linaung: this is a good idea actually.... idk why i so dumb

src/main/java/Uranus/Tasks/Deadlines.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    public String getBy(){
+        return by;
+    }
+
+    @Override
+    public void setDescription(String description){
+        int separatorIndex = description.indexOf(' ');
+        String[] str = description.split("/by");
```
tzerbin: Is it possible to remove the use of magic numbers and strings in the entire codebase?

src/main/java/Uranus/Tasks/Task.java
```diff
@@ -0,0 +1,52 @@
+package Uranus.Tasks;
+
+public class Task {
+    protected String description;
+    protected String commandInput;
+    protected boolean isDone;
+    private String tag;
```
tzerbin: Perhaps be a little more descriptive about the tag?

src/main/java/Uranus/Tasks/Task.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public Task(String description) {
+        this.commandInput = description;
+        setDescription(description);
+        this.isDone = false;
+        this.tag = " "; // for no-tag
```
tzerbin: Be very careful on the usage of an empty String and a String that contains a space character!

src/main/java/Uranus/Utility/FileManagement.java
```diff
@@ -0,0 +1,46 @@
+package Uranus.Utility;
+
+import Uranus.Tasks.Task;
+
+import java.io.*;
+import java.util.Scanner;
+
+public abstract class FileManagement extends Functions{
+
+    public static void save() {
```
tzerbin: Although a nit, but perhaps it would be good to further describe the method e.g. saveFile?
This appears in other parts of the codebase e.g. print instead of printInformation etc.

src/main/java/Uranus/Utility/Functions.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public static void print(String... messages){
+        System.out.println(SEPARATOR);
+        for (String message : messages) {
+            System.out.println(message);
+        }
+        System.out.println(SEPARATOR);
+    }
+
+    public void taskmaster() {
```
tzerbin: Could you please name methods using verbs instead of nouns?

tzerbin (commented):
Good attempt in following the guidelines.
You can work on removing magic strings/numbers usage, and naming of your variables / methods!

PR #12, opened 2024-08-30
[Tan Ting Hui] iP

Ridiculouswifi:


src/main/java/Freedom.java
```diff
@@ -0,0 +1,96 @@
...
+            if (!isCommand) {
+                lastIndex++;
+            }
+        }
+
+        System.out.println(logo + closing + logo);
+    }
+
+    public static boolean handleInput(String input) {
+        String[] words = input.split(" ");
```
xenthm: I like the use of plural form of "word" to indicate an array of words!

src/main/java/Freedom.java
```diff
@@ -0,0 +1,96 @@
...
+            String[] members = description.split("/");
+            storage[lastIndex] = new Deadline(members[0], members[1]);
+        }
+        else if (words[0].equals("event")) {
+            description = input.replaceFirst("event", "");
+            String[] members = description.split("/");
+            storage[lastIndex] = new Event(members[0], members[1], members[2]);
+        } else {
+            storage[lastIndex] = new Task(input);
+        }
```
xenthm: Not sure about this format, maybe you can refer to the Java coding standards document! I think it should be of the form:
```
if (condition) {
    statements;
} else if (condition) {
    statements;
} else {
    statements;
}
```

src/main/java/Freedom.java
```diff
@@ -0,0 +1,96 @@
...
+        System.out.println(logo + message + logo);
+
+        while (true) {
+            Scanner in = new Scanner(System.in);
+            String line = in.nextLine();
+
+            if (line.equals("bye")) {
+                break;
+            }
+            boolean isCommand = handleInput(line);
```
xenthm: I like the use of `is` in the boolean `isCommand`! It makes it sound like a boolean.

src/main/java/Freedom.java
```diff
@@ -0,0 +1,96 @@
+import java.util.Scanner;
+
+public class Freedom {
+    protected static Task[] storage = new Task[100];
```
xenthm: I like that you attached the array specifier `[]` to the array type `Task` and not the array variable `storage`!

xenthm (commented):
I think you did a good job with the coding standards overall! There is just one minor standard violation, but it doesn't change the readability of the code that much.

src/main/java/Freedom.java
tkhahns: I would suggest using `command` instead of `isCommand` for the input variable. You can  always check if the `command` is empty later on, it is just considering it as a boolean may not be the best option

src/main/java/Freedom.java
```diff
@@ -0,0 +1,96 @@
...
+            return true;
+        }
+        if (words[0].equals("todo")) {
+            description = input.replaceFirst("todo", "");
+            storage[lastIndex] = new ToDo(description);
+        }
+        else if (words[0].equals("deadline")) {
+            description = input.replaceFirst("deadline", "");
+            String[] members = description.split("/");
+            storage[lastIndex] = new Deadline(members[0], members[1]);
```
Paulifyer: Perhaps it would be better to use a switch statement instead of using a long set if else statements for clarity.
Additionally the code standard is to use:
if {
.....
} else {
....
}

for if else statements

src/main/java/Event.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public Event(String description, String from, String to) {
+        super(description);
+        setFrom(from.replaceFirst("from ", ""));
+        setTo(to.replaceFirst("to ", ""));
+    }
+
+    public void setFrom(String from) {
+        this.from = from;
+    }
```
Paulifyer: Generally consistent in following naming conventions throughout the code

src/main/java/Freedom.java
```diff
@@ -0,0 +1,96 @@
...
+        if (words[0].equals("unmark")) {
+            int listNumber = Integer.parseInt(words[1]);
+            storage[listNumber - 1].markUndone();
+            markTaskStatement(storage[listNumber - 1], false);
+            return true;
+        }
+        if (words[0].equals("todo")) {
+            description = input.replaceFirst("todo", "");
+            storage[lastIndex] = new ToDo(description);
+        }
```
tkhahns, line 29: consider using `switch` statements for a complex logic structure

src/main/java/Freedom.java
```diff
@@ -0,0 +1,96 @@
+import java.util.Scanner;
+
+public class Freedom {
+    protected static Task[] storage = new Task[100];
+    static int lastIndex = 0;
+
+    public static void main(String[] args) {
+        String logo = "\t_________________________________________\n";
```
Paulifyer: Maybe use the logo variable as a constant instead of a variable?

src/main/java/Deadline.java
tkhahns, line 1: Very good overall, efficient and follows good coding standards. Commit messages are clear and follow good structures

Paulifyer (commented):
Generally okay across the board good work

src/main/java/freedom/tasks/Event.java
```diff
@@ -0,0 +1,62 @@
+package freedom.tasks;
+
+import freedom.exceptions.DescriptionEmpty;
+import freedom.exceptions.TimeEmpty;
+
+public class Event extends Task{
+    protected String from;
+    protected String to;
```
TJW0911: Perhaps using more specific names such as startTime and endTime might be better here.

src/main/java/freedom/tasks/Deadline.java
```diff
@@ -0,0 +1,51 @@
+package freedom.tasks;
+
+import freedom.exceptions.DescriptionEmpty;
+import freedom.exceptions.TimeEmpty;
+
+public class Deadline extends Task{
+    protected String doneBy;
```
TJW0911: Perhaps a more descriptive name such as dueDate/dueTime is better here.

src/main/java/Freedom.java
```diff
@@ -0,0 +1,256 @@
...
+
+            saveData();
+        } catch (Exception e) {
+            System.out.print("");
+        }
+
+        System.out.println(LOGO + CLOSING_MESSAGE + LOGO);
+    }
+
+    public static void handleInput(String input) {
```
thoriritu: In handleInput, adding tasks (todo, deadline, event) is repeated multiple times. You can abstract the task-creation logic into a helper method.

src/main/java/Freedom.java
```diff
@@ -0,0 +1,256 @@
...
+
+        try {
+            checkData();
+            loadData();
+
+            // Super loop for getting inputs
+            while (in.hasNextLine()) {
+                String line = in.nextLine();
+
+                if (line.equals("bye")) {
```
thoriritu: The string "bye" is used multiple times. It can be defined as a constant.

src/main/java/Freedom.java
```diff
@@ -0,0 +1,256 @@
...
+        System.out.print(LOGO + "\tHere are the tasks in your list:\n");
+        for (int i = 0; i < lastIndex; i++) {
+            System.out.println("\t" + counter + "." + tasks.get(i).printLine());
+            counter++;
+        }
+        System.out.println(LOGO);
+    }
+
+    public static void editTask(String[] words) {
+        final int COMMAND_INDEX = 0;
```
thoriritu: Avoid magic numbers. Instead of hardcoding indices like 0, 1, 2 in editTask, declare constants to improve readability.

src/main/java/Freedom.java
thoriritu: Methods like handleInput and editTask are quite large and handle too much responsibility. Consider splitting them into smaller methods.

PR #13, opened 2024-08-31
[TPH777] iP

TPH777:


src/main/java/Iris.java
```diff
@@ -0,0 +1,143 @@
...
+            case "event":
+                newTask = new Event(details);
+                break;
+            default:
+                System.out.println("Invalid Command");
+                return;
+            }
+            tasks[numOfTasks++] = newTask;
+            printAddTaskMessage(newTask);
+        } catch (RuntimeException e) { // For missing description
```
CheeKiangg: I like that you added a comment to improve the readability of what exception was being thrown.

src/main/java/Task.java
```diff
@@ -0,0 +1,51 @@
...
+		case 2:
+			this.date = "("
+					+ formatDate(textParts[1])
+					+ ")";
+			break;
+		case 3:
+			this.date = "("
+					+ formatDate(textParts[1])
+					+ formatDate(textParts[2])
+					+ ")";
```
CheeKiangg: you might want to add  "//fallthrough" comment at the end of case 3

src/main/java/Task.java
```diff
@@ -0,0 +1,51 @@
...
+		}
```
CheeKiangg: perhaps you could also add in a default statement in the event that all switch cases is false.

src/main/java/Iris.java
CheeKiangg: perhaps you could try abstracting addTask function into 2 smaller functions to improve readability.

CheeKiangg (commented):
Overall, good job so far. 
Some minor changes requested. Some of the classes are getting a bit long. Consider splitting them into smaller classes if possible.

src/main/java/Iris.java
```diff
@@ -0,0 +1,143 @@
...
+            taskToUnmark.unmarkFromDone();
+            System.out.println("OK, I've marked this task as not done yet:\n"
+                    + taskToUnmark);
+        }
+        catch (IllegalArgumentException e) {
+            System.out.println(e.getMessage());
+        }
+    }
+
+    private static Task getTaskToEdit(String text) {
```
YukeeHong: Good use of exception handling to catch missing and invalid task numbers. The code makes it clear when the user provides incomplete information.

src/main/java/Iris.java
```diff
@@ -0,0 +1,143 @@
...
+
+    private static void printAddTaskMessage(Task newTask) {
+        System.out.println("Got it. I've added this task:\n"
+                + newTask
+                + "\nNow you have "
+                + numOfTasks
+                + " tasks in the list");
+    }
+
+    public static void markTask(String text) {
```
YukeeHong: To avoid repetition in markTask and unmarkTask, consider creating a helper function that handles both marking and unmarking tasks.

src/main/java/Iris.java
```diff
@@ -0,0 +1,143 @@
+import java.util.Scanner;
+
+public class Iris {
+    public static Task[] tasks = new Task[100];
```
YukeeHong: For better scalability and flexibility, consider using ArrayList<Task> instead of a fixed-size array. This will eliminate the need to check if the array is full and allows dynamic resizing when adding tasks.

src/main/java/Task.java
```diff
@@ -0,0 +1,51 @@
+public class Task {
```
YukeeHong: The task creation logic could be simplified further by encapsulating task type creation into separate factory methods, reducing the size of the addTask method and improving code readability.

src/main/java/Iris.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.ArrayList;
+
+public class Iris {
+    static ArrayList<Task> tasks;
+
+    public static void printDivider() {
+        System.out.println("---------------------------------------------");
+    }
+
+    public static boolean chat(String text) {
```
tzerbin: It would be good to leave comments on why this method is returning a boolean.
We will discuss comments in Week 7.

tzerbin: Perhaps name the method with a better verb?

src/main/java/Iris.java
```diff
@@ -0,0 +1,56 @@
...
+        String command = text.split(" ")[0].toLowerCase();
+        printDivider();
+        switch (command) {
+        case "":
+            System.out.println("Tell me your needs! I'm here to help!");
+            break;
+        case "bye":
```
tzerbin: Would it be possible to remove the need of magic strings and numbers in your entire code base?

src/main/java/exceptions/IrisExceptions.java
```diff
@@ -0,0 +1,7 @@
+package exceptions;
+
+public class IrisExceptions extends Exception {
```
tzerbin: Is there a reason why the class is named in the plural form?

tzerbin (commented):
Good attempt in following the guidelines!
You would want to remove magic strings and numbers in your codebase, together with changing some nits on method and variable naming.

PR #15, opened 2024-08-31
[Tng Wen Xi] iP

wx-03:


src/main/java/Luke.java
```diff
@@ -0,0 +1,193 @@
...
+        } else if (command.equalsIgnoreCase("unmark")) {
+            executeCommand(CommandType.UNMARK, inputArr);
+        } else if (command.equalsIgnoreCase("todo")){
+            executeCommand(CommandType.TODO, inputArr);
+        } else if (command.equalsIgnoreCase("deadline")){
+            executeCommand(CommandType.DEADLINE, inputArr);
+        } else if (command.equalsIgnoreCase("event")){
+            executeCommand(CommandType.EVENT, inputArr);
+        }
+    }
```
Bev-low: Is there a better way to do this other than nested else if?

src/main/java/CommandType.java
```diff
@@ -0,0 +1,9 @@
+public enum CommandType {
```
Bev-low: Good use of Enums here, makes the code quite clean

src/main/java/Luke.java
```diff
@@ -0,0 +1,193 @@
...
+                }
+            }
+            if (fromIdx == -1) {
+                printReply("From when???");
+                break;
+            }
+            if (toIdx == -1) {
+                printReply("To when???");
+                break;
+            }
```
Bev-low: These if statments can be taken out and replaces with Enums

src/main/java/Luke.java
```diff
@@ -0,0 +1,193 @@
...
+            tasks[idx].setAsDone();
+            type = tasks[idx].getClass();
+            t = 'U';
+            if (type.equals(Deadline.class)) {
+                t = 'D';
+            } else if (type.equals(ToDo.class)) {
+                t = 'T';
+            } else if (type.equals(Event.class)) {
+                t = 'E';
+            }
```
Bev-low: These if statements also can be removed and replace with Enums

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,301 @@
...
+import java.io.FileNotFoundException;
+import java.io.FileWriter;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Scanner;
+
+public class Luke {
+
+    private static final String LOGO =
```
tzerbin: Very good work in using constants to prevent the use of magic strings and numbers!

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,301 @@
...
+        String line;
+        for (Task t : tasks) {
+            line = t.toString2();
+            fw.write(line);
+            fw.write(System.lineSeparator());
+        }
+        fw.close();
+    }
+
+    private static void executeCommand(CommandType commandType, String[] inputArr) {
```
tzerbin: This is a very long method. Perhaps extract functionalities out for each command so as to allow your code to be more readable?

tzerbin: Would you want to name your parameter as `inputArr` instead of `inputs`?

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,301 @@
...
+        default:
+            throw new InvalidCommand("Invalid command");
+        }
+    }
+
+    private static void sendMessage(String userInput) {
+        String[] inputArr = userInput.split(" ");
+        String command = inputArr[0];
+
+        if (command.equalsIgnoreCase("bye")) {
```
tzerbin: This is a very long piece of code. You may want to explore inheritance together with polymorphism to cut down the amount of code here.

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,301 @@
...
+        printDivider();
+        System.out.println(LOGO);
+        System.out.println("Hi im Luke!\nWhat can I do for you? :)");
+        printDivider();
+    }
+
+    private static void loadSingleTask(String taskStr) {
+        String[] taskStrArr = taskStr.split("\\|");
+        switch (taskStrArr[TASK_TYPE_INDEX]) {
+        case "T" -> tasks.add(new ToDo(taskStrArr[DESCRIPTION_INDEX], taskStrArr[ISDONE_INDEX].equals("1")));
```
tzerbin: To prevent error-prone code, perhaps you should use colons ':' instead of arrow "->" instead?
You can also use default branches.

src/main/java/luke/Luke.java
```diff
@@ -0,0 +1,301 @@
...
+                taskStrArr[ISDONE_INDEX].equals("1")));
+        }
+    }
+
+    private static void loadSavedTasks() throws IOException {
+        File dataDir = new File("data");
+        if (!dataDir.exists()) {
+            dataDir.mkdir();
+        }
+        File save = new File("data/tasks.txt");
```
tzerbin: There are still quite a number of magic strings lying around the codebase. Please take a look!

src/main/java/luke/exceptions/IncorrectInput.java
```diff
@@ -0,0 +1,7 @@
+package luke.exceptions;
+
+public class IncorrectInput extends LukeException {
```
tzerbin: You may want to name your classes with a postfix "Exception" that are extending from LukeException?

src/main/java/luke/tasks/Deadline.java
```diff
@@ -0,0 +1,30 @@
+package luke.tasks;
+
+public class Deadline extends Task{
+    protected String by;
+
```
tzerbin, line 5: There is an excessive use of whitespace here.
(This is part of the intermediate portion of coding standard.)

src/main/java/luke/tasks/Event.java
```diff
@@ -0,0 +1,35 @@
...
+    }
+    public String getTo() {
+        return to;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[E][%s] %s (from: %s to: %s)", this.isDone ? "X" : " ", this.description, from, to);
+    }
+    public String toString2() {
```
tzerbin: Could you rename the method with a more descriptive name, instead of labelling methods with integer postfix?

tzerbin (commented):
Good attempt in following the guidelines, especially on the attempt of removing magic strings.
Some nits here and there!

PR #18, opened 2024-08-31
[Nigel Yeo] iP

NigelYeoTW:


src/main/java/Event.java
```diff
@@ -0,0 +1,22 @@
+public class Event extends Task{
+    protected String eventStart;
+    protected String eventEnd;
+
+    public Event(String input) {
+        super(input.substring(input.indexOf(" ") + 1, input.indexOf("/from") - 1));
+        int startingIndexOfEventStart = input.indexOf("/from") + 6;
+        int endingIndexOfEventStart = input.indexOf("/to") - 1;
+        int startingIndexOfEventEnd = endingIndexOfEventStart + 5;
```
chewycj: I like how you named these variables, it helps me understand how this part of the code works :)

src/main/java/Legin.java
```diff
@@ -0,0 +1,125 @@
...
+                addTodo(input);
+                break;
+            case "deadline":
+                addDeadline(input);
+                break;
+            case "event":
+                addEvent(input);
+                break;
+            default:
+                echo(input);
```
chewycj: Should there be a break; after echo(input)?

src/main/java/Legin.java
```diff
@@ -0,0 +1,125 @@
...
+                break;
+            case "todo":
+                addTodo(input);
+                break;
+            case "deadline":
+                addDeadline(input);
+                break;
+            case "event":
+                addEvent(input);
+                break;
```
chewycj: This part of the code is really clean, with good indentations! :D

src/main/java/Legin.java
```diff
@@ -0,0 +1,125 @@
...
+        System.out.println("Hello! I'm Legin, your best online companion!");
+        System.out.println("What can I do for you today my friend :D");
+        horizontalLine();
+    }
+    public static void bye() {
+        horizontalLine();
+        System.out.println("Bye " +
+            Character.toString(0x1F44B) +
+            ". Hope to see you again really soon! " +
+            Character.toString(0x1F608));
```
chewycj: Should this indentation for wrapped lines be 8 spaces instead of 4 spaces?

```suggestion
        System.out.println("Bye " +
                Character.toString(0x1F44B) +
                ". Hope to see you again really soon! " +
                Character.toString(0x1F608));
```

chewycj (commented):
Overall, I found your code easy to read! There are two changes to take note of, one for line wrapping and another for the default case in the switch statement 👍

src/main/java/Legin.java
```diff
@@ -0,0 +1,125 @@
+import java.util.Scanner;
+
+public class Legin {
+    private static Task[] tasks = new Task[100];
+    private static int currentTaskCount = 0;
+
+    public static void horizontalLine() {
+        System.out.println("______________________" +
+                "______________________________________");
+    }
```
crystal-bys: good to do this! as one might want to print lines for every output, doing this is logical so you can call it when needed and keep things neat.

src/main/java/Legin.java
```diff
@@ -0,0 +1,125 @@
...
+
+    private static void printAddedTaskMessage() {
+        horizontalLine();
+        System.out.println("Got it. I've added this task:");
+        System.out.println(tasks[currentTaskCount]);
+        currentTaskCount++;
+        System.out.println("Now you have " + currentTaskCount + " tasks in the list.");
+        horizontalLine();
+    }
```
crystal-bys: good job doing this as it prevents repeated lines of code! I see that you call this quite often below :)

src/main/java/Legin.java
```diff
@@ -0,0 +1,125 @@
...
+                echo(input);
+            }
+        }
+    }
+
+    public static void main(String[] args) {
+        greet();
+        runBot();
+        bye();
+    }
```
crystal-bys: I was surprised when I see this, good to see that all you managed to keep things simple here

src/main/java/Legin.java
crystal-bys: I didn't think of doing a default case. good job!

crystal-bys (commented):
I like how the code is kept simple and neat down the code, and I am inspired to do the same for mine. Great job!

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
+import exception.LeginMissingParamsException;
+import exception.LeginEmptyTaskException;
+import exception.LeginIndexOutOfBoundsException;
+import exception.LeginInvalidCommandException;
+import task.Deadline;
+import task.Event;
+import task.Task;
+import task.Todo;
+import java.io.IOException;
+import java.util.ArrayList;
```
tzerbin: Good job on providing explicit imports. Perhaps you would want to order them in a collated order e.g. java.util.ArrayList is together with java.util.Scanner, since both comes from java.util package.
Note that this is part of Java coding standard - Intermediate.

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
...
+import java.io.File;
+import java.io.FileWriter;
+import java.util.Scanner;
+
+public class Legin {
+    private static final ArrayList<Task> tasks = new ArrayList<>();
+    private static int currentTaskCount = 0;
+    private static final String FILE_PATH = "./tasklist.txt";
+
+    public static void horizontalLine() {
```
tzerbin: You would want to use verbs to name a method instead of nouns.
```suggestion
    public static void printHorizontalLine() {
```

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
...
+        System.out.println("______________________" +
+                "______________________________________");
+    }
+
+    public static void addTextDataToArray(String[] words) throws LeginEmptyTaskException {
+        String taskType = words[0];
+        switch(taskType) {
+        case "T":
```
tzerbin: Could you prevent the usage of magic numbers and characters?

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
...
+        currentTaskCount++;
+    }
+
+    public static void inputTextFileData() {
+        try {
+            Scanner s = new Scanner(new File(FILE_PATH));
+            while (s.hasNext()) {
+                String[] words = s.nextLine().split("\\|");
+                addTextDataToArray(words);
+            }
```
tzerbin: Please look out for indentations.

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
...
+    public static void addEvent(String input) {
+        try {
+            tasks.add(new Event(input));
+            printAddedTaskMessage();
+        } catch (LeginEmptyTaskException | LeginMissingParamsException e) {
+            printExceptionMessage(e);
+        }
+    }
+
+    public static void list() {
```
tzerbin: Perhaps be more specific on the naming of this method.

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
...
+                throw new LeginInvalidCommandException();
+            }
+        } catch (LeginInvalidCommandException | LeginIndexOutOfBoundsException e) {
+            printExceptionMessage(e);
+        }
+        return false;
+    }
+
+    public static void runBot() {
+        boolean saidBye = false;
```
tzerbin: Perhaps name the boolean with a prefix "has"?

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    public static void greet() {
+        horizontalLine();
+        System.out.println("Hello! I'm Legin, your best online companion!");
+        System.out.println("What can I do for you today my friend :D");
+        inputTextFileData();
+        horizontalLine();
+    }
+    public static void bye() throws IOException {
```
tzerbin: Similar to a previous issue, would you like to use verbs for this method instead?

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,45 @@
...
+public class Deadline extends Task{
+    protected String by;
+
+    private static String getDeadlineDescription(String input) throws LeginMissingParamsException,
+            LeginEmptyTaskException {
+        validityCheck(input);
+        return input.substring(input.indexOf(" ") + 1, input.indexOf("/by") - 1);
+    }
+
+    private static void validityCheck(String input) throws LeginMissingParamsException,
```
tzerbin: You would want to use a verb here instead.
```suggestion
    private static void checkValidity(String input) throws LeginMissingParamsException,
```

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,45 @@
...
+            LeginEmptyTaskException {
+        int indexOfBy = input.indexOf("/by");
+        if (indexOfBy == 9) {
```
tzerbin: Would you be able to remove the magic numbers in this file (as well as other files)?

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,45 @@
...
+        }
+    }
+
+    public Deadline(String input) throws LeginEmptyTaskException, LeginMissingParamsException {
+        super(getDeadlineDescription(input));
+        int startingIndexOfDueDate = input.indexOf("/by") + 4;
+        this.by = input.substring(startingIndexOfDueDate);
+    }
+
+    public Deadline(String description, String by) throws LeginEmptyTaskException {
```
tzerbin: What does "by" refers to? It would be good if this is named more appropriately.

tzerbin (commented):
It is a very good attempt to follow all the guidelines.
Do make sure you get rid of magic strings / numbers and change some of the naming for methods and boolean variables.

src/main/java/Legin.java
```diff
@@ -0,0 +1,232 @@
...
+        switch(taskType) {
+        case "T":
+            tasks.add(new Todo(words[2], true));
+            break;
+        case "D":
+            tasks.add(new Deadline(words[2], words[3]));
+            break;
+        case "E":
+            tasks.add(new Event(words[2], words[3], words[4]));
+            break;
```
tzerbin: You may want to prevent error-prone code by adding default branch.

PR #22, opened 2024-08-31
[Jaden Lim] iP

jadenlimjc:


src/main/java/ListUtils.java
```diff
@@ -0,0 +1,155 @@
...
+            super(description);
+            this.by = by;
+        }
+        @Override
+        public String toString() {
+            return "[D]" + super.toString() + " (by: " + by + ")";
+        }
+    }
+
+    private static class Event extends Task {
```
frederickemerson: I really love the implementation, but do you think creating a new class would make it more cleaner rather than putting Tasks as part of List Utils

src/main/java/ListUtils.java
```diff
@@ -0,0 +1,155 @@
...
+            return "[E]" + super.toString() + " (from: " + from + ", to: " + to + ")";
+        }
+
+    }
+    // create list of tasks
+    private static List<Task> userInputs = new ArrayList<>();
+
+    //create methods to add each respective task subclasses
+
+    public static void addTodoToList(String description) {
```
frederickemerson: do you feel like you are repeating a lot of lines, for adding to the list, is there a workaround this

src/main/java/Quag.java
```diff
@@ -0,0 +1,150 @@
+import java.util.Scanner;
+import java.util.Arrays;
+
+public class Quag {
+    public static void main(String[] args) {
```
frederickemerson, line 15: LGTM, the main looks quite clean

src/main/java/Quag.java
```diff
@@ -0,0 +1,150 @@
...
+                    break;
+            }
+        }
+    }
+
+    private static String getCommand(String userInput) {
+        //find first space
+        int firstSpaceIndex = userInput.indexOf(" ");
+        // first word is the command
+        if (firstSpaceIndex != -1) {
```
frederickemerson: Love the implementation here

src/main/java/EchoUtils.java
```diff
@@ -0,0 +1,7 @@
+public class EchoUtils {
+    public static void echo(String input) {
```
frederickemerson: ooh looks quite clean, but do you think you need a whole class for a single function

src/main/java/Quag.java
```diff
@@ -0,0 +1,150 @@
...
+            String from = userInput.substring(fromIndex + fromLength, toIndex).trim();
+            String to = userInput.substring(toIndex + toLength).trim();
+            ListUtils.addEventToList(description, from, to);
+        }  else {
+            System.out.println("That's not a valid quaggin format for an event!!");
+        }
+
+    }
+
+    private static void displayCommandList() {
```
MayFairMI6: Here, comment should be self explanatory. For the description part, one can change the function name todisplayCommandListWithDescription or comment can be shortened to just #commands with description (preferably the function name)

src/main/java/Quag.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private static void addEvent(String userInput) {
+        //description is from parts[1] until first /
+        //from is in between first and second /
+        //to is after second /
+
+        // int to denote length and position of "/from" and "/to" separator
+        int fromIndex = userInput.indexOf("/from");
+        int fromLength = 6;
```
MayFairMI6: Avoided magic numbers effectively, enhancing clarity

src/main/java/Quag.java
```diff
@@ -0,0 +1,150 @@
...
+                    addTodo(userInput);
+                    break;
+                case "deadline":
+                    addDeadline(userInput);
+                    break;
+                case "event":
+                    addEvent(userInput);
+                    break;
+                default:
+                    displayCommandList();
```
MayFairMI6: Effective use of default statement to handle unexpected inputs which would otherwise confuse the user

src/main/java/Quag.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    private static void addTodo(String userInput) {
+        //description is from parts[1] to end
+        String description = userInput.substring("todo".length()).trim();
+        ListUtils.addTodoToList(description);
+    }
+
+    private static void addDeadline(String userInput) {
+        //description is from parts[1] until / is detected
```
MayFairMI6: Code should be as modular as possible, and global variables should be used minimally (including references in the comments). It would be difficult to always trace to a previous function to understand what parts [1] means.

MayFairMI6 (commented):
Code looks good with no deeply nested loops or high cyclomatic complexity however there are some flaws concerning naming conventions and modularity which I have highlighted. Thanks!

src/main/java/ListUtils.java
```diff
@@ -0,0 +1,244 @@
...
+
+        @Override
+        public String toString() {
+            return "[T]" + super.toString();
+        }
+    }
+
+    private static class Deadline extends Task {
+
+        protected String by;
```
TJW0911: Perhaps it would be a good idea to use a name such as dueDate for clarity as by could be an ambiguous name.

src/main/java/ListUtils.java
```diff
@@ -0,0 +1,244 @@
...
+        }
+        @Override
+        public String toString() {
+            return "[D]" + super.toString() + " (by: " + by + ")";
+        }
+    }
+
+    private static class Event extends Task {
+        protected String from;
+        protected String to;
```
TJW0911: Perhaps the variable names could be improved since they are referring to a specific time. You could consider using startTime and endTime.

src/main/java/ListUtils.java
```diff
@@ -0,0 +1,244 @@
...
+                dir.mkdirs();
+            }
+            FileWriter writer = new FileWriter(FILE_PATH);
+            for (Task task : userInputs) {
+                if (task instanceof Todo) {
+                    writer.write("T | " + (task.isDone ? "1" : "0") + " | " + task.description + "\n");
+                } else if (task instanceof Deadline) {
+                    writer.write("D | " + (task.isDone ? "1" : "0") + " | " + task.description + " | " + ((Deadline) task).by + "\n");
+                } else if (task instanceof Event) {
+                    writer.write("E | " + (task.isDone ? "1" : "0") + " | " + task.description + " | " + ((Event) task).from + " | " + ((Event) task).to + "\n");
```
TJW0911: Maybe it would be a good idea to refactor and create a method for each of these if statements to improve code readability.

src/main/java/ListUtils.java
TJW0911, line 1: Generally it might be a good idea to put task, deadline, event and todo in their own separate files instead of putting all of them in ListUtils.java for better readability.

PR #23, opened 2024-08-31
[yeekian] iP

yeekian:


src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
+public class List {
+    public static final String INVALID_EVENT_INPUT_MESSAGE = "event <event name> /from <start date/time> /end <end date/time>";
```
RCPilot1604: I like the extraction of every string constant!

src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
...
+    public static final int TODO_WORD_LEN = 4;
+    public static final String DEADLINE_BY_KEYWORD = "/by";
+    public static final String EVENT_FROM_KEYWORD = "/from";
+    public static final String EVENT_TO_KEYWORD = "/to";
+
+    private int numItems;
+    private Task[] itemList = new Task[100];
+
+    public List() {
+        numItems = 0;
```
RCPilot1604: Would it be better to use "this.numItems = 0" instead?

src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    public int getNumItems() {
+        return numItems;
+    }
+
+    public void addItem(String line) {
+        if (line.length() >= 7 && line.substring(0, 5).equals("event")) {
```
RCPilot1604: Shouldn't it be extracted out instead of a magic number?

src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
...
+            if (line.contains(EVENT_FROM_KEYWORD) && line.contains(EVENT_TO_KEYWORD)) {
+                addEvent(line);
+            } else {
+                printInvalidEventMessage();
+            }
+        } else if (line.length() >= (DEADLINE_WORD_LEN + INPUT_SPACE_BUFFER) && line.substring(0, DEADLINE_WORD_LEN).equals("deadline")) {
```
RCPilot1604: Would it be better to break up this long expression into small, more bite-sized steps?

src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
...
+            eventStartDate = line.substring(indexOfStartDatePrefix, indexOfEndDatePrefix).trim().replace("/from ", "");
+        }
+
+        return eventStartDate;
+    }
+
+    public void printList() {
+        System.out.println("\tHere are the tasks in your list:");
+        for (int i = 0; i < numItems; i++) {
+            System.out.println("\t" + (i + 1) + "." + itemList[i]);
```
RCPilot1604: I like the use of the toString() override to allow easier printing of the elements in itemList<Task>.

RCPilot1604 (commented):
Nice job! :)

src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
...
+                printTaskMarkedMessage(itemNum);
+            }
+        } catch (Exception e) {
+            printInputIndexNotAnIntegerMessage();
+        }
+    }
+
+    private void printTaskMarkedMessage(int itemNum) {
+        System.out.println("\tNice! I've marked this task as done:");
+        System.out.println("\t  " + itemList[itemNum -1]);
```
tsoudibi: a small thing, need a space between - and 1

src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
...
+                printTaskUnmarkedMessage(itemNum);
+            }
+        } catch (Exception e) {
+            printInputIndexNotAnIntegerMessage();
+        }
+    }
+
+    private void printTaskUnmarkedMessage(int itemNum) {
+        System.out.println("\tOK, I've marked this task as not done yet:");
+        System.out.println("\t  " + itemList[itemNum -1]);
```
tsoudibi: still need space between - and 1

src/main/java/Mel.java
```diff
@@ -0,0 +1,68 @@
+import java.util.Scanner;
+
+public class Mel {
+
+    public static final String DRAW_HORIZONTAL_LINE = "\t________________________________________";
+    public static final int MARK_WORD_LEN = 4;
+    public static final int UNMARK_WORD_LEN = 6;
+    public static final int INPUT_SPACE_BUFFER = 2;
```
tsoudibi: what is the purpose of INPUT_SPACE_BUFFER?

src/main/java/List.java
```diff
@@ -0,0 +1,221 @@
+public class List {
+    public static final String INVALID_EVENT_INPUT_MESSAGE = "event <event name> /from <start date/time> /end <end date/time>";
+    public static final String INVALID_DEADLINE_INPUT_MESSAGE = "deadline <deadline name> /by <deadline>";
+    public static final String INVALID_TODO_INPUT_MESSAGE = "todo <task name>";
+    public static final int DEADLINE_WORD_LEN = 8;
+    public static final int INPUT_SPACE_BUFFER = 2;
+    public static final int TODO_WORD_LEN = 4;
```
tsoudibi: maybe it would be better to use `input.contains("something")` rather than using the length of the comment. Anyway it still works

tsoudibi (commented):
LGTM. but it would be better to add more comment in you code.

src/main/java/Mel.java
yeekian: The purpose of the buffer is the space and first digit that follows the commands

src/main/java/List.java
yeekian: Thank you for your suggestion, the toString() function is already being called in itemList[i].

src/main/java/List.java
yeekian: Done thank you

src/main/java/List.java
```diff
@@ -0,0 +1,315 @@
...
+    }
+
+    private static void printInvalidTaskMessage() {
+        System.out.println("\tInvalid command format:" + System.lineSeparator() + "\t\t" + INVALID_TODO_INPUT_MESSAGE
+                + System.lineSeparator() + "\t\t" + INVALID_DEADLINE_INPUT_MESSAGE + System.lineSeparator() + "\t\t"
+                + INVALID_EVENT_INPUT_MESSAGE + System.lineSeparator() + "\t\t"
+                + INVALID_MARK_MESSAGE + System.lineSeparator() + "\t\t"
+                + INVALID_UNMARK_MESSAGE + System.lineSeparator() + "\t\t"
+                + INVALID_DELETE_MESSAGE);
+    }
```
Mahesh1772: Consider using switch case, as the different values that you are expecting are predetermined

src/main/java/List.java
```diff
@@ -0,0 +1,315 @@
...
+        if (indexOfStartDatePrefix > indexOfEndDatePrefix) {
+            eventStartDate = line.substring(indexOfStartDatePrefix).replaceFirst("/from", "").trim();
+        } else {
+            eventStartDate = line.substring(indexOfStartDatePrefix, indexOfEndDatePrefix).replaceFirst("/from", "").trim();
+        }
+
+        dateFieldNotEmpty(eventStartDate);
+
+        return eventStartDate;
+    }
```
Mahesh1772: Consider having fewer methods to reduce the number of lines of duplicated code

src/main/java/List.java
```diff
@@ -0,0 +1,315 @@
...
+
+    private void printTaskMarkedMessage(int itemNum) {
+        System.out.println("\tNice! I've marked this task as done:");
+        System.out.println("\t  " + itemArrayList.get(itemNum - 1));
+    }
+
+    private static void printInputIndexOutOfRangeMessage() {
+        System.out.println("\tInput index number out of range.");
+    }
```
Mahesh1772: Consider making these strings into constants as well

src/main/java/List.java
```diff
@@ -0,0 +1,315 @@
...
+            } else {
+                this.markListItemAsUnDone(itemNum);
+                printTaskUnmarkedMessage(itemNum);
+            }
+        } catch (NumberFormatException  e) {
+            printInputIndexNotAnIntegerMessage();
+        } catch (Exception e) {
+            printUnknownErrorMessage();
+        }
+    }
```
Mahesh1772: Consider catching specific errors inorder to prompt the user with more informative text error. Consider `ArrayIndexOutOfBoundsException`, etc.

src/main/java/Mel.java
```diff
@@ -0,0 +1,204 @@
...
+            System.out.println("File not found: " + filePath);
+        }
+    }
+
+    private static void parseAndAddItem(String line, List userList) {
+        String[] parts = line.split(" \\| ");
+        String taskType = parts[0]; // T, D, E
+        boolean isDone = parts[1].equals("X");
+        String taskDescription = parts[2];
+
```
Mahesh1772: Add checks to ensure that the number of fields matches what’s expected. If not, log an error and skip the line.

src/main/java/Mel.java
```diff
@@ -0,0 +1,204 @@
...
+            } else if (isDelete(line)) {
+                userList.deleteItem(line);
+                printHorizontalLine();
+                saveListToFile(userList);
+
+            } else {
+                userList.addItem(line);
+                printHorizontalLine();
+                saveListToFile(userList);
+
```
Mahesh1772: Consider using a switch case here instead of using many methods and if-else cases. This could be harder to maintain and does not look as clean

src/main/java/task/Deadline.java
```diff
@@ -0,0 +1,22 @@
+package task;
+
+public class Deadline extends Task {
+
+    protected String by;
```
Mahesh1772: Consider naming it more informative such as `byDate`

src/main/java/task/Task.java
```diff
@@ -0,0 +1,31 @@
...
+
+    public void markAsDone() {
+        isDone = true;
+    }
+
+    public void markAsUnDone() {
+        isDone = false;
+    }
+
+    public String formattedTask() {
```
Mahesh1772, line 28: Good practice to define here and override in all the sub-classes

Mahesh1772 (commented):
The overall implementation is solid, but I recommend a few improvements. Please ensure consistent and descriptive naming for variables and methods, refactor if-else blocks into switch statements where applicable for clarity, and add Javadoc comments for all classes and methods to improve documentation. These changes will enhance code readability and maintainability.

PR #29, opened 2024-09-01
[Goh Eng Hui, Jeremy] iP

jemehgoh:


src/main/java/Nell.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+    }
+
+    /**
+     * Adds a new task to the task list
+     *
+     * @param taskToAdd The task to be added to the list
+     */
+    private static void listAddTask(Task taskToAdd) {
+        // Stores text in task list
```
ChuaCleon: redundant comment here, easily understandable 2 lines of code

src/main/java/Nell.java
```diff
@@ -0,0 +1,217 @@
...
+        Task taskToAdd = new Task(command);
+        listAddTask(taskToAdd);
+        System.out.printf("-> added: %s%n", command);
+    }
+
+    /**
+     * Executes a bye command
+     *
+     */
+    private static void executeCommandBye() {
```
ChuaCleon: good naming convention, however I would suggest you to remove "executeCommand" from your function names (ie just leaving the command name as the function name). This way it improves readability of the code

src/main/java/Nell.java
```diff
@@ -0,0 +1,217 @@
...
+            System.out.println("-> Invalid task!");
+        }
+    }
+
+    /**
+     * Executes an add task command (no command word) with a given command body
+     *
+     * @param command The command body
+     */
+    private static void executeCommandAddTask(String command) {
```
ChuaCleon: Instead of including the execute functions in the Nell class, might want to consider creating a separate class to include these functionalities like (CommandManager or something) to delegate command management and improve abstraction in your code this way

src/main/java/Nell.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Executes a deadline command with a given command body
+     *
+     * @param commandBody The command body
+     */
+    private static void executeCommandDeadline(String commandBody) {
+        System.out.println("-> The task has been added to the list:");
+        String[] commandWords = commandBody.split("/by");
+        Deadline deadlineToAdd = new Deadline(commandWords[0].trim(), commandWords[1].trim());
```
ChuaCleon: good and simple implementation, easily understandable code

src/main/java/Nell.java
jemehgoh: This makes sense, but for some of these commands (like "todo" and "deadline"), I believe that following your suggestion would make the method name potentially misleading, and also go against naming conventions (method names being verbs). Would it be better for me to change the method names to "execute<command>" instead?

src/main/java/Nell.java
jemehgoh: I agree with this, thanks for your suggestion!

src/main/java/Nell.java
```diff
@@ -0,0 +1,217 @@
...
+        tasks[taskCount] = taskToAdd;
+        taskCount++;
+    }
+
+    /**
+     * Lists out the currently stored tasks in TaskList, upon receipt of the
+     * list command
+     *
+     */
+    private static void executeCommandList() {
```
Yisoong: I noticed that "executeCommand" is repeated a lot in the naming of your functions and it might be better to shorten the name to focus on what the function is doing.

src/main/java/Nell.java
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * Executes an unmark command with a given command body
+     *
+     * @param commandBody The command body
+     */
+    private static void executeCommandUnmark(String commandBody) {
+        int taskIndex = Integer.parseInt(commandBody);
+        if (checkIfTaskValid(taskIndex)) {
+            // Marks a task as not done
```
Yisoong: This comment is redundant.

src/main/java/Nell.java
```diff
@@ -0,0 +1,217 @@
...
+        System.out.println("   " + deadlineToAdd);
+        System.out.println(String.format("   The list now has %d tasks", taskCount));
+    }
+
+    /**
+     * Executes an event command with a given command body
+     *
+     * @param commandBody The command body
+     */
+    private static void executeCommandEvent(String commandBody) {
```
Yisoong: commandBody sounds a bit clunky and confusing. It might be better to call it "description" for clarity.

src/main/java/Event.java
```diff
@@ -0,0 +1,15 @@
...
+    private String to;
+
+    public Event(String description, String from, String to) {
+        super(description);
+        this.from = from;
+        this.to = to;
+    }
+
+    public String toString() {
+        return String.format("[E][%s] %s (from: %s to: %s)", this.getStatusIcon(),
```
Yisoong: This line is confusing and seems to be doing a lot in one line. Perhaps it would be better to break it up or use another way to achieve this result.

Yisoong (commented):
Good job with your IP! I think some of the naming conventions could be made clearer but overall, great code! Jiayous!!

src/main/java/Nell.java
jemehgoh: Thanks for your comment, I agree with your suggestion.

src/main/java/Event.java
jemehgoh: I'm personally not sure of how to go about this. Perhaps I could use a taskType variable (in the parent class), and move some of the functionality into the parent class?

src/main/java/nell/tasks/Task.java
```diff
@@ -0,0 +1,68 @@
...
+     * done or not
+     *
+     * @return X if isDone is true, space otherwise
+     */
+    public char getStatusIcon() {
+        if (isDone) {
+            return 'X';
+        }
+
+        return ' ';
```
Mahesh1772: Consider using ternary operator, to for simpler readability.

src/main/java/nell/tasks/Task.java
```diff
@@ -0,0 +1,68 @@
...
+     *
+     * @param done The boolean value isDone is to be set to
+     */
+    public void setDone(boolean done) {
+        isDone = done;
+    }
+
+    @Override
+    public String toString() {
+        return String.format("[%S][%s] %s", this.type, this.getStatusIcon(), this.description);
```
Mahesh1772, line 63: Good use of string formatting!

src/main/java/nell/tasks/ToDo.java
```diff
@@ -0,0 +1,17 @@
...
+public class ToDo extends nell.tasks.Task {
+
+    public ToDo(String description) {
+        super(description, "T");
+    }
+
+    public ToDo(String description, boolean isDone) {
+        super(description, isDone);
+        this.type = "T";
+    }
```
Mahesh1772: Consider adding Javadocs for these constructors as well.

src/main/java/nell/tasks/Task.java
```diff
@@ -0,0 +1,68 @@
+package nell.tasks;
+
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    protected String type;
```
Mahesh1772, line 11: Consider changing 'type' to a final, if it will not be modified.

src/main/java/nell/tasks/Event.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public String toString() {
+        return String.format("%s (from: %s to: %s)", super.toString(),
+                this.from, this.to);
+    }
+
+    public String getFileLine() {
+        return String.format("%s|%s|%s", super.getFileLine(), this.from, this.to);
+    }
+}
```
Mahesh1772: Consider adding JavaDocs

src/main/java/nell/Nell.java
```diff
@@ -0,0 +1,346 @@
...
+        boolean isGettingCommands = true;
+
+        while (isGettingCommands) {
+            // Get user command and respond accordingly
+            String command = input.nextLine();
+            String[] commandWords = command.split(" ", 2);
+            switch (commandWords[0]) {
+            case "bye":
+                sayBye();
+                saveToFile();
```
Mahesh1772: Try to save tasks at the end of each operation, so that even when the app is not closed properly it retains the changes made.

src/main/java/nell/Nell.java
```diff
@@ -0,0 +1,346 @@
...
+    private static void loadTask(String taskLine) {
+        String[] taskParameters = taskLine.split("\\|");
+        String taskType = taskParameters[0];
+        String taskDescription = taskParameters[2];
+
+        boolean taskIsDone = false;
+        if (taskParameters[1].equals("X"))
+        {
+            taskIsDone = true;
+        }
```
Mahesh1772: Could be simplified to one line: boolean taskIsDone = taskParameters[1].equals("X");

Mahesh1772 (commented):
The code is clean, follows conventions, and is well-structured. I’ve provided a few suggestions mainly to improve clarity, conciseness, and documentation—nothing critical. 

Consider making the methods in the main file shorter for better readability.

Your attention to detail and consistent coding practices are evident. Keep up the fantastic effort!

PR #31, opened 2024-09-01
[LuxLucky7] iP

LuxLucky7:


src/main/java/Main.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public Deadlines createDeadlineTask(String enteredString) {
+        String taskDescription=enteredString.substring("deadline".length(), enteredString.indexOf('/')-1);
+        String deadlineTime=enteredString.substring(enteredString.indexOf("/by")+"/by".length()+1);
+        return new Deadlines(taskDescription, deadlineTime);
+    }
+
+    public Events createEventTask(String enteredString) {
+        String taskDescription=enteredString.substring("event".length(), enteredString.indexOf('/')-1);
```
G13nd0n: the event.length() perhaps can be a constant variable instead of using the method

src/main/java/Deadlines.java
```diff
@@ -0,0 +1,19 @@
+public class Deadlines extends Task{
+    private String by;
```
G13nd0n: Perhaps can put as protected instead of private, so that the member of the same package can access its values instead of using a getter

src/main/java/Main.java
```diff
@@ -0,0 +1,133 @@
...
+        for(Task task : tasks) {
+            serialNumber++;
+            System.out.println(serialNumber + "." + task.toString());
+        }
+        printHorizontalLine();
+    }
+
+
+    public void changeTaskStatus(String enteredString) {
+        int taskNumber= Integer.parseInt(enteredString.substring(enteredString.length()-1));
```
G13nd0n: Perhaps can break it down abit instead of all in a line to allow more readability

src/main/java/Main.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public ToDos createTodoTask(String enteredString) {
+        String taskDescription=enteredString.substring("todo".length());
+        return new ToDos(taskDescription);
+    }
+    public void addNewTask(String enteredString) {
+        Task newTask;
+
+        if(enteredString.startsWith("deadline")) {
```
G13nd0n: Is it possible to use switch-case instead of if-else

src/main/java/Main.java
```diff
@@ -71,15 +71,15 @@ public void markTask(Task task) {
     }
 
     public Deadlines createDeadlineTask(String enteredString) {
-        String taskDescription=enteredString.substring("deadline".length(), enteredString.indexOf('/'));
-        String deadlineTime=enteredString.substring(enteredString.indexOf("/by")+1);
+        String taskDescription=enteredString.substring("deadline".length(), enteredString.indexOf('/')-1);
+        String deadlineTime=enteredString.substring(enteredString.indexOf("/by")+"/by".length()+1);
```
mikolajed: Maybe it would a good idea to make these lines shorter

src/main/java/Main.java
```diff
@@ -71,15 +71,15 @@ public void markTask(Task task) {
...
         return new Deadlines(taskDescription, deadlineTime);
     }
 
     public Events createEventTask(String enteredString) {
-        String taskDescription=enteredString.substring("event".length(), enteredString.indexOf('/'));
-        String eventFromTime=enteredString.substring(enteredString.indexOf("/from")+1, enteredString.indexOf("/to"));
-        String eventToTime=enteredString.substring(enteredString.indexOf("/to")+1);
+        String taskDescription=enteredString.substring("event".length(), enteredString.indexOf('/')-1);
+        String eventFromTime=enteredString.substring(enteredString.indexOf("/from")+"/from".length()+1, enteredString.indexOf("/to"));
+        String eventToTime=enteredString.substring(enteredString.indexOf("/to")+"/to".length()+1);
```
mikolajed: It looks too long, maybe I would be better to use more abstraction

src/main/java/Main.java
```diff
@@ -77,19 +62,48 @@ public void changeTaskStatus(String enteredString) {
...
+        } else if(enteredString.startsWith("event")) {
+            newTask=createEventTask(enteredString);
+        } else if(enteredString.startsWith("todo")) {
+            newTask=createTodoTask(enteredString);
+        } else {
+            System.out.println("Invalid task type. Please try again.");
+            printHorizontalLine();
+            return;
+        }
+
```
mikolajed: You followed well the standard for if-else statements

src/main/java/Edith.java
```diff
@@ -19,10 +20,18 @@ public void talkToUser() {
             if(temporyString.equals("bye")){
```
mikolajed: I have noticed no space between if and condition, here and a few other places. You might try look for it in the next commits.

mikolajed (commented):
LGTM. Code follows coding standards with some exceptions.

src/main/java/Deadlines.java
```diff
@@ -0,0 +1,19 @@
+public class Deadlines extends Task{
```
tzerbin: Classes could be wrapped within packages.

tzerbin: Is there a reason why the class name is in plural form?
This is evident in some other classes as well.

src/main/java/Main.java
```diff
@@ -0,0 +1,144 @@
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class Main {
+    Scanner sc=new Scanner(System.in);
+    private final String myName;
```
tzerbin: Constants should be named in SCREAMING_SNAKE_CASE.

src/main/java/Main.java
```diff
@@ -0,0 +1,144 @@
...
+    ArrayList<Task> tasks;
+
+    public Main(String myName) {
+        this.myName = myName;
+        tasks=new ArrayList<>();
+    }
+
+    public void printHorizontalLine() {
+        System.out.println("______________________________________________");
```
tzerbin: Please refrain from using magic strings and numbers throughout the codebase.

src/main/java/Main.java
```diff
@@ -0,0 +1,144 @@
...
+            return;
+        }
+
+        tasks.add(newTask);
+        System.out.println("Got it. I've added this task: ");
+        System.out.println(newTask);
+        System.out.println("Now you have " + tasks.size() + " tasks in the list");
+        printHorizontalLine();
+    }
+    public void talkToUser() {
```
tzerbin: Perhaps use a better verb instead of "talk" to describe the method?

src/main/java/Main.java
```diff
@@ -0,0 +1,144 @@
...
+            } else if(enteredString.contains("mark")) {
+                changeTaskStatus(enteredString);
+            } else {
+                addNewTask(enteredString);
+            }
+        } while (true);
+    }
+
+    public static void main(String[] args) {
+        Main Edith = new Main("Edith");
```
tzerbin: Variable names should not start with a capital letter.

src/main/java/TaskType.java
```diff
@@ -0,0 +1,5 @@
+public class TaskType extends Exception {
```
tzerbin: The class name does not reflect well as an exception class. It sounded more like a class that objects can be instantiated instead.

src/main/java/ToDos.java
```diff
@@ -0,0 +1,9 @@
+public class ToDos extends Task{
```
tzerbin: Is there a reason why this class is named in the plural form?

tzerbin (commented):
Good attempt on following guidelines!

There are a few crucial issues such as naming classes in plural form, not naming variables appropriately, not wrapping classes within packages, and usage of magic strings & numbers.

PR #33, opened 2024-09-01
[iaso1774] iP

iaso1774:


src/main/java/Gertrude.java
DucPhong135, line 1: This code is a bit hard to read, maybe divide into different methods for each of the case will be better, avoid writing everything in a main method. 
Too many if-else statements maybe hard for debugging issue or for other to read your code.
Variable names can be more meaningful like the lineArr variable, it hard to know what does this variable contain

Overall, good code, just need some small improvement

DucPhong135 (commented):
Overall, the code is good!

src/main/java/Gertrude.java
```diff
@@ -0,0 +1,286 @@
...
+    public static void addTodo(String[] lineInputArr, ArrayList<Task> tasks) {
+        String name = "";
+        for(int i = 1; i < lineInputArr.length; i++) {
+            name += lineInputArr[i] + " ";
+        }
+        Todo newTodo = new Todo(name);
+        addTask(newTodo, name, tasks);
+    }
+
+    public static void addDeadline(String[] lineInputArr, ArrayList<Task> tasks) throws GertrudeException, KeywordException {
```
thoriritu: The methods addDeadline and addEvent share similar logic (such as iterating through the array and concatenating strings). This can be refactored to avoid duplication

src/main/java/Gertrude.java
```diff
@@ -0,0 +1,286 @@
...
+        if(end.isEmpty()) {
+            System.out.println("To is empty.");
+            throw new GertrudeException();
+        }
+
+        Event newEvent = new Event(description, start, end);
+        addTask(newEvent, description, tasks);
+    }
+
+    public static ArrayList<Task> openFile() throws IOException, CorruptedFileException, IndexOutOfBoundsException {
```
thoriritu: You could have a FileManager class that handles the file reading and writing independently of the logic in Gertrude. This improves code reusability and reduces the responsibilities of each class.

src/main/java/Gertrude.java
```diff
@@ -0,0 +1,286 @@
...
+        }
+        if(deadline.isEmpty()) {
+            System.out.println("Deadline is empty.");
+            throw new GertrudeException();
+        }
+
+        Deadline newDeadline = new Deadline(description, deadline);
+        addTask(newDeadline, description, tasks);
+    }
+
```
thoriritu, line 39: Add more comments to make the code more readable and understandable.

PR #42, opened 2024-09-01
[AndrewNguyen4] iP

AndrewNguyen4:


gu0y1:
Observation: The array `Task tasks[] = new Task[100];` is defined with square brackets after the variable name.

Suggestion: It is recommended to use the format `Task[] tasks = new Task[100];` for array definitions. This aligns with the Java coding standard where the array brackets should be placed next to the type, not the variable.

src/main/java/Aerus.java
```diff
@@ -0,0 +1,72 @@
...
+                }
+                System.out.println(DIVIDER_LINE);
+                continue;
+            }
+
+            // Case: Mark & Unmark
+            String[] userInputSplit = userInput.split(" ");
+
+            // Test if the input is formatted like a mark/unmark command
+            if (userInputSplit.length == 2 && userInputSplit[1].matches("\\d+(\\.\\d+)?") &&
```
ThienDuc3112: Avoid complicated expression

Can split into different part

src/main/java/Aerus.java
```diff
@@ -0,0 +1,72 @@
...
+                    printContent("I have unmarked this task:\n\t" + tasks[taskNumber].toString());
+                    continue;
+                }
+            }
+
+            // Case: Add task
+            if (!userInput.isEmpty()) {
+                tasks[Task.tasksCount] = new Task(userInput);
+                printContent("Added task: " + userInput);
+            }
```
ThienDuc3112: Practice KISSing

Could've just use `userInput.startsWith` to check

src/main/java/Task.java
```diff
@@ -0,0 +1,21 @@
+public class Task {
+    protected String description;
+    protected boolean isDone;
+    protected static int tasksCount = 0;
```
ThienDuc3112: Structure code logically

`tasksCount` should be in the same place as Task array, may cause confusion to reader

src/main/java/Aerus.java
```diff
@@ -0,0 +1,72 @@
+import java.util.Scanner;
+
+import static java.lang.Integer.parseInt;
+
+public class Aerus {
+
+    public static final String DIVIDER_LINE = "____________________________________________________________";
```
ThienDuc3112: good use of final

ThienDuc3112 (approved):
Overall lgtm

src/main/java/Aerus.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        String userInput;
+        Task tasks[] = new Task[100];
+
+        printContent("Hello! I'm Aerus!\nWhat can I do for you?");
+
+        // while block handling Aerus' main functionality
+        while (true) {
```
JuvenLimHY: use switch case instead of while loop

src/main/java/Aerus.java
```diff
@@ -0,0 +1,72 @@
...
+            // Case: Terminate program
+            if (userInput.equals("bye")) {
+                break;
+            }
+
+            // Case: See list
+            if (userInput.equals("list")) {
+                if (Task.tasksCount == 0) {
+                    printContent("You don't have any tasks!");
+                    continue;
```
JuvenLimHY: instead of using continue, change to a flag to break the while loop.

src/main/java/Aerus.java
```diff
@@ -0,0 +1,72 @@
...
+import static java.lang.Integer.parseInt;
+
+public class Aerus {
+
+    public static final String DIVIDER_LINE = "____________________________________________________________";
+
+    public static void main(String[] args) {
+        Scanner scanner = new Scanner(System.in);
+        String userInput;
+        Task tasks[] = new Task[100];
```
JuvenLimHY: Avoid magic number

src/main/java/Aerus.java
JuvenLimHY: Avoid complicated expressions

these conditions are too complicated and long. these things can be done before

JuvenLimHY (commented):
LGTM but can be better

src/main/java/Event.java
```diff
@@ -0,0 +1,40 @@
...
+        this.from = from;
+        this.to = to;
+    }
+
+    public static void createNewEvent(String userInput) throws InvalidCreateEventException {
+        String[] eventInfo = userInput.split("/from|/to");
+        if (eventInfo.length == 3) {
+            Event event = new Event(eventInfo[0], eventInfo[1], eventInfo[2]);
+            Aerus.tasks[Task.tasksCount-1] = event;
+            UI.printContent("Added Event: " + event.toString());
```
okkhoy: some of this code can be extracted to a parser class

src/main/java/InputHandler.java
```diff
@@ -0,0 +1,65 @@
...
+                if (taskNumber > Task.tasksCount) {
+                    throw new InvalidMarkException();
+                } else if (userInputSplit[0].equals("mark")) {
+                    Aerus.tasks[taskNumber].isDone = true;
+                    Aerus.tasks[taskNumber].printMark();
+                    return 1;
+                } else if (userInputSplit[0].equals("unmark")) {
+                    Aerus.tasks[taskNumber].isDone = false;
+                    Aerus.tasks[taskNumber].printUnmark();
+                    return 1;
```
okkhoy: these can be extracted as execute methods in each command class.

src/main/java/Aerus.java
```diff
@@ -0,0 +1,30 @@
+import java.util.Scanner;
+
+import static java.lang.Integer.parseInt;
+
+public class Aerus {
+
+    protected static Task[] tasks = new Task[100];
```
okkhoy: how about expanding the task list?

okkhoy (commented):
please take a look at the comments and complete the code where necessary

PR #44, opened 2024-09-01
[Sukkaito] iP

Sukkaito:


src/main/java/SuBOT.java
```diff
@@ -0,0 +1,73 @@
+import java.util.Scanner;
+
+public class SuBOT {
+    private static boolean stopFlag = false;
```
rharwo:
```suggestion
    private static boolean stopFlag = isRunning;
```

src/main/java/SuBOT.java
```diff
@@ -0,0 +1,73 @@
...
+    private static Task[] tasks = new Task[100];
```
rharwo: Is it wise to use magic numbers? Perhaps you could make it a final variable.

src/main/java/SuBOT.java
```diff
@@ -0,0 +1,73 @@
...
+    private static int taskCount;
+
+    /**
+     * Process user's input command
+     * @param command Input command
+     */
+    public static void process(String command) {
+        String[] argv = command.split(" ");
+        int argc = argv.length;
```
rharwo: Perhaps you could use more descriptive names than argv and argc.

src/main/java/SuBOT.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public static void main(String[] args) {
+        String logo = """
+                       ____  ____  ______
+           _______  __/ __ )/ __ \\/_  __/
+          / ___/ / / / __  / / / / / /
+         (__  ) /_/ / /_/ / /_/ / / /
+        /____/\\__,_/_____/\\____/ /_/
+        """;
```
rharwo: Perhaps change the logo to a final variable if it is not going to change.

src/main/java/SuBOT.java
```diff
@@ -0,0 +1,73 @@
...
+
+        System.out.println(logo);
+        System.out.println("How can I help you today?");
+        System.out.println("-------------------------------");
```
rharwo: Could you use a final variable so that you do not have to keep repeating "----" in your code?

rharwo (commented):
Looks good overall, some minor code quality fixes.

src/main/java/SuBOT.java
Sukkaito: Thank you for the suggestion! Added! 👍

src/main/java/SuBOT.java
Sukkaito: Thank you for the suggestion! Added! 👍

src/main/java/SuBOT.java
Sukkaito: Thank you for the suggestion! Added! 👍

src/main/java/SuBOT.java
Sukkaito: Thank you for the suggestion! Added! 👍

src/main/java/SuBOT.java
Sukkaito: I think `argc` and `argv` are already standard names here (at least in C++). It shouldn't be a problem.

src/main/java/ActionManager.java
```diff
@@ -0,0 +1,163 @@
...
+            int taskIndex = Parser.parseMarkUnmark(argString);
+            if (taskIndex >= taskCount) {
+                System.out.println("Task not found");
+                return;
+            }
+            boolean isMark = action.equals("mark");
+            tasks[taskIndex].setDone(isMark);
+            System.out.printf("I have %sed the following task for you:\n", action);
+            tasks[taskIndex].printTask();
+        } catch (SuBotException e) {
```
Kobot7: Is the variable name "e" informative enough?

src/main/java/ActionManager.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+
+        for (int i = 0; i < taskCount; ++i) {
+            System.out.printf("%d.", i + 1);
+            tasks[i].printTask();
+        }
+    }
+
+    /**
+     * Marks/unmarks a task as done
```
Kobot7: Leave an empty line between description and parameter section

src/main/java/ActionManager.java
```diff
@@ -0,0 +1,163 @@
...
+     * @param action <code>mark</code> or <code>unmark</code>
```
Kobot7: Ensure that there is punctuation behind each parameter description

src/main/java/ActionManager.java
```diff
@@ -0,0 +1,163 @@
...
+        }
+        try {
+            switch (action) {
+            case "bye":
+                SuBOT.isRunning = false; //Halt signal for parent class
+                break;
+            case "list":
+                listAllTask();
+                break;
+            case "mark":
```
Kobot7: Is there meant to be a "break" here? Otherwise, it would be good to indicate that this was intended using "//Fallthrough"

src/main/java/Deadline.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public boolean isDone() {
+        return super.isDone();
+    }
+
+    @Override
+    public void setDone(boolean done) {
+        super.setDone(done);
+    }
```
Kobot7: Inherited methods do not need to be redeclared if there are no changes to their functionality

src/main/java/Parser.java
```diff
@@ -0,0 +1,92 @@
...
+        if (fromSlashIndex == -1)  {
+            throw new SuBotException("\"/from\" expected");
+        }
+
+        int toSlashIndex = args.indexOf("/to");
+        if (toSlashIndex == -1)  {
+            throw new SuBotException("\"/to\" expected");
+        }
+
+        String eventName = args.substring(0, fromSlashIndex).strip();
```
Kobot7: Consider using a named constant instead of a magic number

src/main/java/Parser.java
```diff
@@ -0,0 +1,92 @@
...
+     */
+    static int parseMarkUnmark(String args) throws SuBotException {
+        if (args.isBlank()) {
+            throw new SuBotException("Task number cannot be blank");
+        }
+        try {
+            int taskIndex = Integer.parseUnsignedInt(args.trim()) - 1; //Index starts from 0
+            if (taskIndex < 0) {
+                throw new SuBotException("Task number must be a positive integer");
+            } else { return taskIndex; }
```
Kobot7: Follow the layout convention for if-else statements, format it using Egyptian style

PR #62, opened 2024-09-02
[ctzeyong] iP

ctzeyong:


src/main/java/Melchizedek.java
```diff
@@ -0,0 +1,44 @@
+import java.util.Scanner;
+
+public class Melchizedek {
+
+    public static TaskList taskList = new TaskList();
+
+    public static void printHorizontalLine() {
```
Chinorea: You could improve the naming of the function to something more specific, in the event you change the functionality of this.

You can consider renaming it to printParition or printBorder, in the event the border is change to vertical line or a line of + signs

src/main/java/Melchizedek.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static void greetUser() {
+        printHorizontalLine();
+        System.out.println("\tHello! I'm Melchizedek.");
+        System.out.println("\tHow can I be a blessing to you?");
+        printHorizontalLine();
+    }
+
+    public static void sayByeToUser() {
```
Chinorea: Good naming convention of methods (), function written in camelCase and is in verb form

src/main/java/Melchizedek.java
```diff
@@ -0,0 +1,44 @@
+import java.util.Scanner;
+
+public class Melchizedek {
+
+    public static TaskList taskList = new TaskList();
+
+    public static void printHorizontalLine() {
+        System.out.println("\t____________________________________________________________");
```
Chinorea: Consider declaring as a constant by using public static final String :.....

src/main/java/TaskList.java
```diff
@@ -0,0 +1,51 @@
+public class TaskList {
+    private Task[] allTasks;
+    private int noOfTasks;
+
+    public TaskList(String task){
+        allTasks = new Task[100];
+        allTasks[0] = new Task(1, task);
+        noOfTasks = 1;
```
Chinorea: Consider renaming the variable to something more intuitive such as numberOfTasks instead of abreviating number as no

src/main/java/melchizedek/Melchizedek.java
```diff
@@ -0,0 +1,150 @@
...
+        System.out.println("\t  to unmark a task as done: unmark TASK_NUMBER");
+        System.out.println("\t  to display all tasks on the list: list");
+        System.out.println("\t  to exit: bye");
+    }
+
+    public static void main(String[] args) {
+        sayHelloToUser();
+        FileHandling.loadFile(taskList);
+        Scanner in = new Scanner(System.in);
+        while (true) {
```
thomasjlalba, line 27: Maybe the looping logic could be extracted out to avoid long methods?

src/main/java/melchizedek/Melchizedek.java
```diff
@@ -0,0 +1,150 @@
...
+            String input = in.nextLine();
+            String[] tokens = input.split(" ");
+
+            printSeparator();
+
+            switch (tokens[0].toLowerCase()) {
```
thomasjlalba, line 32: Maybe `tokens[0]` could be refactored to make it clearer that it is the command? Same for `tokens[1]` and a few others!

src/main/java/melchizedek/Melchizedek.java
```diff
@@ -0,0 +1,150 @@
...
+                    System.out.println("\tExample: deadline coding assignment /by 12pm");
+                }
+                break;
+
+            case "event":
+                try {
+                    if (tokens.length < 2) {
+                        throw new DescriptionNotPresentException();
+                    }
+                    taskList.addEvent(Arrays.copyOfRange(tokens, 1, tokens.length));
```
thomasjlalba, line 99: Maybe it would be good to refactor this magic literal `1` that occurs in several places!

thomasjlalba (commented):
Some refactoring could be done on magic literals and long methods, but well done on following the coding standards and quality.

PR #66, opened 2024-09-02
[rahuljai-05] iP

rahuljai-05:


src/main/java/Fenix.java
```diff
@@ -0,0 +1,124 @@
...
+                        "Please provide a valid task number to mark");
+            }
+            break;
+        case "unmark":
+            if (words.length > 1 && isValidIndex(words[1])) {
+                unmarkTaskAsDone(Integer.parseInt(words[1]) - 1);
+            } else {
+                System.out.println(
+                        "Please provide a valid task number to unmark");
+            }
```
sarahchow03: Cases "mark" and "unmark" can be abstracted into mark and unmark functions.

src/main/java/Fenix.java
```diff
@@ -0,0 +1,124 @@
+import java.util.Scanner;
+
+public class Fenix {
+
+    public static final int HORIZONTAL_LINE_USER_COMMAND_LENGTH = 60;
+    public static final int HORIZONTAL_LINE_FENIX_MODIFICATION_LENGTH = 52;
```
sarahchow03: Good use of constants to prevent the use of magic numbers in the code base. Excellent job!

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
...
+        return taskName;
+    }
+
+    public boolean isDone() {
+        return isDone;
+    }
+
+    public void markAsDone() {
+        this.isDone = true;
+    }
```
sarahchow03, line 74: Output of markTaskAsDone() in main can be placed in this function. Similarly, this can be done for unmarkAsDone().

sarahchow03 (commented):
Overall, function names and variables have been appropriately named. You could work on the documentation of code through the use of comments, so that it is more readable.

Good job!

src/main/java/Task.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public void unmarkAsDone() {
+        this.isDone = false;
+    }
+
+    // Override
```
sarahchow03, line 26: nit: Annotation of Override is incorrect and should be @Override instead.

src/main/java/Fenix.java
```diff
@@ -0,0 +1,124 @@
...
+        return scanner.nextLine();
+    }
+
+    private void processUserInput(String userInput) {
+        String[] words = userInput.trim().split(" ");
+        String command = words[0];
+        switch (command) {
+        case "bye":
+            bidFarewell();
+            return;
```
iaso1774: If there isn't a break statement, there is supposed to be a //Fallthrough comment, although I'm not sure if this applies if you instead have a return statement.

src/main/java/Fenix.java
```diff
@@ -0,0 +1,124 @@
...
+
+    public void bidFarewell() {
+        System.out.println(FAREWELL);
+        System.out.println(HORIZONTAL_LINE_USER_COMMAND);
+        scanner.close();
+    }
+
+    public void showAllTasks(boolean isModified) {
+        String modifiedString = (isModified ? "\t" : "");
+        for (int i = 0; i < taskArray.length && taskArray[i] != null; i += 1) {
```
iaso1774: Use i++ instead of i += 1

src/main/java/Fenix.java
```diff
@@ -0,0 +1,124 @@
+import java.util.Scanner;
+
+public class Fenix {
+
```
iaso1774: There shouldn't be a blank line here.

iaso1774 (commented):
I could only find a few minor coding standard errors. Nice job!!!

src/main/java/Fenix.java
```diff
@@ -0,0 +1,17 @@
+public class Fenix {
+    public static void main(String[] args) {
+        String horizontalLine = "_".repeat(60);
+        String greeting = "Hello I'm Fenix";
+        String request = "What can I do for you ?";
+        String farewell = "Bye. Hope to see you again soon!";
+        System.out.println(horizontalLine);
+        System.out.println(greeting + "\n" + request);
```
DucPhong135: Should avoid writing everything on a main method, divide into different method for easy manage.
There are no spaces so it is hard to keep up with the workflow.
Try to not using the \n in your print statement, instead use another print statement to show that those two are on different lines.

DucPhong135 (commented):
Overall, since you have't finish yet, I cannot give further comments on your work

src/main/java/Event.java
```diff
@@ -0,0 +1,23 @@
+public class Event extends Deadline {
+    protected String deadlineTill;
```
TJW0911, line 6: Perhaps it would be a good idea to name this something like "endTime" since events generally have a start and end time instead of using deadlineTill.

src/main/java/Event.java
```diff
@@ -0,0 +1,23 @@
...
+    public String getDeadlineTill(String[] inputArray) {
+        return inputArray[2];
+    }
+
+    @Override
+    public String toString() {
+        String taskStatus = "[E][" + (isDone ? "X" : " ") + "] ";
+        String from = " (from: " + deadline.replaceFirst("^from\\s+", "");
+        String to = "to: " + deadlineTill.replaceFirst("^to\\s+", "") + ")";
+        return taskStatus + taskName + from + to;
```
TJW0911: Using deadline and deadlineTill to represent the start and end time seems a bit confusing for the reader. Maybe it would be a good idea to have Event extend Task instead.

src/main/java/Fenix.java
```diff
@@ -0,0 +1,197 @@
...
+    private boolean isValidTaskNumber(String input) {
+        try {
+            int index = Integer.parseInt(input);
+            return index > 0 && index <= taskNumber;
+        } catch (NumberFormatException e) {
+            return false;
+        }
+    }
+
+    private void unmarkAsDone(String[] words) {
```
TJW0911: unmarkAsDone sounds a bit confusing, perhaps you could use markTaskAsUndone instead.

src/main/java/Fenix.java
```diff
@@ -0,0 +1,197 @@
...
+
+    public void markTaskAsDone(int taskNumber) {
+        System.out.println("Task successfully completed. A job well executed.");
+        taskArray[taskNumber].markAsDone();
+        System.out.println(HORIZONTAL_LINE_FENIX_MODIFICATION);
+        showAllTasks(true);
+        System.out.println(HORIZONTAL_LINE_FENIX_MODIFICATION);
+    }
+
+    public void unmarkTaskAsDone(int taskNumber) {
```
TJW0911: Similarly unmarkTaskAsDone sounds a bit confusing, perhaps you could use markTaskAsUndone instead.

PR #71, opened 2024-09-02
[louisjoety] iP

louisjoety:


src/main/java/Echo.java
```diff
@@ -0,0 +1,118 @@
+import java.util.Scanner;
+
+public class Echo {
+    private static final String SEPARATOR = "____________________________________________________________";
```
KuanHsienn: Good use of constants but may consider using "_".repeat(30) to improve conciseness

src/main/java/Echo.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private static void addTask(String userInput, TaskList taskList) {
+        userInput = userInput.trim();
+
+        if (userInput.isEmpty()) {
+            return;
+        }
+
+        if (userInput.startsWith("todo ")) {
+            String description = userInput.substring(5).trim();
```
KuanHsienn: Using of magic numbers should be avoided, can replace it with a constant such as descIndex

src/main/java/Echo.java
```diff
@@ -0,0 +1,118 @@
...
+            ToDo newTask = new ToDo(description);
+            taskList.storeTask(newTask);
+            System.out.println("____________________________________________________________");
+            System.out.println("Got it. I've added this task:");
+            System.out.println(newTask);
+            System.out.println("Now you have " + taskList.getTaskNumber() + " tasks in the list.");
+            System.out.println("____________________________________________________________");
```
KuanHsienn: These lines are repeated in your if-else branches. You can actually pull it outside to shorten the code and prevent code repetition

src/main/java/TaskList.java
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public String markTaskAsNotDone(int index) {
+        if (index < 1 || index > taskNumber) {
+            return "Invalid task number.";
+        }
+        Task task = tasks[index - 1];
+        task.setAsUndone();
+        return "OK, I've marked this task as not done yet:\n" + task;
+    }
+}
```
KuanHsienn: Overall good code and can work to reduce code repetition

wongwh2002: Coding standards is generally is good, naming convention is easily understood, just have to reduce the use of magic numbers.

src/main/java/Echo.java
```diff
@@ -0,0 +1,118 @@
...
+        System.out.println(SEPARATOR);
+        System.out.println(message);
+        System.out.println(SEPARATOR);
+    }
+
+    private static void printTaskList(TaskList taskList) {
+        System.out.println(SEPARATOR);
+        System.out.println("Here are the tasks in your list:");
+        System.out.println(taskList.displayTasks());
+        System.out.println(SEPARATOR);
```
wongwh2002: Is code is repeated for many of the handle functions, eg for the printing?

src/main/java/Echo.java
```diff
@@ -0,0 +1,118 @@
...
+            taskList.storeTask(newTask);
+            System.out.println("____________________________________________________________");
+            System.out.println("Got it. I've added this task:");
+            System.out.println(newTask);
+            System.out.println("Now you have " + taskList.getTaskNumber() + " tasks in the list.");
+            System.out.println("____________________________________________________________");
+        } else if (userInput.startsWith("deadline ")) {
+            int byIndex = userInput.indexOf(" /by ");
+            String description = userInput.substring(9, byIndex).trim();
+            String dueDate = userInput.substring(byIndex + 5).trim();
```
wongwh2002: Use of magic numbers might be confusing

src/main/java/Echo.java
```diff
@@ -0,0 +1,118 @@
...
+            String description = userInput.substring(6, fromIndex).trim();
+            String start = userInput.substring(fromIndex + 6, toIndex).trim();
+            String end = userInput.substring(toIndex + 4).trim();
+            Event newTask = new Event(description, start, end);
+            taskList.storeTask(newTask);
+            System.out.println("____________________________________________________________");
+            System.out.println("Got it. I've added this task:");
+            System.out.println(newTask);
+            System.out.println("Now you have " + taskList.getTaskNumber() + " tasks in the list.");
+            System.out.println("____________________________________________________________");
```
wongwh2002: you already have separator constant, is there a reason to use println ("______") here

src/main/java/Echo.java
louisjoety: Oh nice! That works, thanks for letting me know.

src/main/java/Echo.java
louisjoety: I will use constants for this to make it clearer 👍

src/main/java/Echo.java
louisjoety: Thanks for the feedback. @KuanHsienn

src/main/java/TaskList.java
louisjoety: Thanks for the feedback. @wongwh2002

src/main/java/Echo.java
louisjoety: I will use constants for this to make it clearer 👍

src/main/java/Echo.java
louisjoety: I might have missed it when I was refactoring, thanks for pointing this out.

src/main/java/Command/Command.java
```diff
@@ -0,0 +1,8 @@
+package Command;
+
+import TaskList.TaskList;
+
+public abstract class Command {
+    protected static final String SEPARATOR = "_".repeat(30);
```
okkhoy: see if you can move some other members from child classes here in addition to the separator.

src/main/java/Command/DeadlineCommand.java
```diff
@@ -0,0 +1,37 @@
...
+import Task.Deadline;
+import Exception.EchoException;
+
+public class DeadlineCommand extends Command {
+    private static final int DEADLINE_WORD_LENGTH = 9;
+    private static final int DEADLINE_DUE_DATE_OFFSET = 5;
+
+    @Override
+    public void execute(TaskList taskList, String userInput) {
+        int byIndex = userInput.indexOf(" /by ");
```
okkhoy: can SLAP this out

src/main/java/Command/DeadlineCommand.java
```diff
@@ -0,0 +1,37 @@
+package Command;
+
+import TaskList.TaskList;
+import Task.Deadline;
+import Exception.EchoException;
+
+public class DeadlineCommand extends Command {
+    private static final int DEADLINE_WORD_LENGTH = 9;
```
okkhoy: what if user includes additional spaces?

src/main/java/Command/DeleteCommand.java
```diff
@@ -0,0 +1,28 @@
...
+            if (taskNumber < 1 || taskNumber > taskList.getTaskNumber()) {
+                System.out.println(SEPARATOR);
+                System.out.println(EchoException.taskNumberOutOfRange());
+                System.out.println(SEPARATOR);
+                return;
+            }
+            System.out.println(SEPARATOR);
+            System.out.println(taskList.deleteTask(taskNumber));
+            System.out.println(SEPARATOR);
+        }catch (NumberFormatException e) {
```
okkhoy: spacing problem

src/main/java/Command/EventCommand.java
```diff
@@ -0,0 +1,42 @@
+package Command;
+
+import TaskList.TaskList;
+import Task.Event;
+import Exception.EchoException;
+
+public class EventCommand extends Command {
+    private static final int EVENT_WORD_LENGTH = 6;
+    private static final int START_WORD_LENGTH = 7;
```
okkhoy: same comment as deadline class above - if it is a common issue, pls fix

src/main/java/Echo.java
```diff
@@ -0,0 +1,18 @@
+import Parser.Parser;
```
okkhoy: package names are small letters. so `parser` instead of `Parser`

src/main/java/TaskList/TaskList.java
```diff
@@ -0,0 +1,125 @@
...
+        storage = new Storage();
+        loadTasks();
+    }
+
+    public int getTaskNumber() {
+        return tasks.size();
+    }
+
+    /**
+     * Storage of new task in the task list.
```
okkhoy: incorrect format for header comment

okkhoy (commented):
In general, well written code; pay attention to package names

PR #81, opened 2024-09-02
[isaacsaw25] iP

isaacsaw25:


src/main/tasks/Task.java
```diff
@@ -0,0 +1,73 @@
...
+     * Returns a string representation of the task.
+     * The representation includes the status icon and the task description.
+     *
+     * @return A string representing the task.
+     */
+    @Override
+    public String toString() {
+        return "[" + getStatusIcon() + "] " + description;
+    }
+}
```
wenxin-c: Please leave a few line at the end of each page.

src/main/java/CheonsaBot.java
```diff
@@ -0,0 +1,421 @@
...
+        createFileIfNotExists(); 
+        loadTasksFromFile();    
+        getGreeting();
+        try (Scanner scanner = new Scanner(System.in)) {
+            boolean running = true;
+            while (running) {
+                String userInput = scanner.nextLine();
+                running = parseInput(userInput);
+            }
+        }
```
wenxin-c: You may consider extracting this part as well so that the level of abstraction will be the same.
E.g.
```
readData();
salary = basic * rise + 1000;
tax = (taxable ? salary * 0.07 : 0);
displayResult();
```
versus
```
readData();
processData();
displayResult();

```

src/main/java/CheonsaBot.java
```diff
@@ -0,0 +1,421 @@
...
+                return false;
+            case "list":
+                printTaskList();
+                break;
+            case "delete":
+                deleteTask(argument);
+                break;
+            default:
+                addTask(command, words);
+                break;
```
wenxin-c: There shouldn't be indentation for case clause. 
E.g.
```
switch (condition) {
case ABC:
    statements;
    // Fallthrough
case DEF:
    statements;
    break;
case XYZ:
    statements;
    break;
default:
    statements;
    break;
}

```

src/main/java/CheonsaBot.java
```diff
@@ -0,0 +1,421 @@
...
+     */
+    private static void markTask(String taskNumber) {
+        try {
+            int index = Integer.parseInt(taskNumber) - 1;
+            if (index >= 0 && index < tasks.size()) {
+                tasks.get(index).setAsDone();
+                updateTaskFile();
+                System.out.println(getHorizontalLine());
+                System.out.println("Marked task as done: " + tasks.get(index));
+                System.out.println(getHorizontalLine());
```
wenxin-c: You may want to abstract the printing output part and pass in parameters to differentiate as the similar pattern in used in a few other methods as well.

src/main/java/CheonsaBot.java
```diff
@@ -0,0 +1,421 @@
...
+        }
+    }
+
+    /**
+     * Adds a new task based on the command and argument provided.
+     *
+     * @param command The type of task to add (e.g., "todo", "deadline", "event").
+     * @param words The details of the task.
+     */
+    private static void addTask(String command, String[] words) {
```
wenxin-c: This method is pretty long, you may consider extracting each case into a separate method. It will be better to keep methods short.

src/main/java/CheonsaBot.java
```diff
@@ -0,0 +1,421 @@
...
+                            throw new IllegalArgumentException("Oops! The event format is incorrect or missing :( \nUse: event <description> /from <start> /to <end>");
+                        }
+                        task = new Event(eventDescription, from, to);
+                    } else {
+                        throw new IllegalArgumentException("Oops! Missing description for 'event' :( ");
+                    }
+                    break;
+
+                default:
+                    throw new IllegalArgumentException("Oh no! Unknown task type: " + command + " :(\nTry commands todo, deadline, or event with the correct format!");
```
wenxin-c: You can consider wrapping the sentence if it gets too long.

src/main/java/CheonsaBot.java
```diff
@@ -0,0 +1,421 @@
...
+     * @return A line to be saved to file on disk.
+     */
+    private static String formatTaskForFile(Task task) {
+        String taskType = "";
+        if (task instanceof ToDo) {
+            taskType = "T";
+        } else if (task instanceof Deadline) {
+            taskType = "D";
+        } else if (task instanceof Event) {
+            taskType = "E";
```
wenxin-c: You may consider converting task types (i.e. "T", "D", "E") into named constants. This also applies to other magic literals especially when they are used multiple times in the code.

wenxin-c (commented):
Overall good job. There are some suggestions for you to consider, The ```CheonsaBot``` class is very long. One suggestion is to split this class into multiple smaller class e.g. parser class to pass data.

PR #87, opened 2024-09-02
[naveen42x] iP

naveen42x:


src/main/java/grok/Grok.java
```diff
@@ -0,0 +1,89 @@
...
+                System.out.println("Got it. I've added this task:");
+                System.out.println(tasks[taskCount]);
+                taskCount++;
+                System.out.println("Now you have " + taskCount + (taskCount == 1 ? " task" : " tasks") + " in the list.");
+                printLine();
+            } else {
+                printLine();
+                System.out.println("I'm sorry, I don't understand that command.");
+                printLine();
+            }
```
Progresst-8, line 42: Due to the complex nature of your main function, maintaining it is going to be difficult. Perhaps break it down into smaller methods?

Progresst-8 (commented):
Overall, not much to review other than splitting up main into smaller functions/classes, i.e. some refactoring needed.

src/main/java/grok/Grok.java
```diff
@@ -21,6 +28,7 @@ public static void main(String[] args) {
                 break;
             } else if (input.equals("list")) {
```
kestryix: Would it be better to change this to a switch case instead of using an if else statement?

src/main/java/grok/Grok.java
```diff
@@ -39,11 +47,35 @@ public static void main(String[] args) {
                 System.out.println("OK, I've marked this task as not done yet:");
                 System.out.println(tasks[taskNumber]);
                 printLine();
-            } else {
-                tasks[taskCount] = new Task(input);
+            } else if (input.startsWith("todo ")) {
+                tasks[taskCount] = new Todo(input.substring(5)); // remove "todo " prefix
```
kestryix: Would it be possible to extract common sections into a function and reuse the function?

src/main/java/grok/tasks/Deadline.java
```diff
@@ -0,0 +1,15 @@
+package grok.tasks;
+
+public class Deadline extends Task {
+    protected String by;
+
+    public Deadline(String description, String by) {
```
kestryix, line 6: would it be possible for the class to handle the string splitting instead of the main code?

src/main/java/grok/tasks/Event.java
```diff
@@ -0,0 +1,17 @@
+package grok.tasks;
+
+public class Event extends Task {
```
kestryix, line 3: would it be possible for the class code to handle the necessary logic instead of the main code in Grok.java?

src/main/java/grok/Grok.java
```diff
@@ -1,9 +1,16 @@
...
+import grok.tasks.Deadline;
+import grok.tasks.Event;
+
 import java.util.Scanner;
 
 public class Grok {
     public static void main(String[] args) {
         Scanner scanner = new Scanner(System.in);
-        Task[] tasks = new Task[100]; // Array to store Task objects
+        Task[] tasks = new Task[100];
```
kestryix: would it be possible to make the magic number a named constant?

src/main/java/grok/Grok.java
naveen42x: Yes, good idea

src/main/java/grok/Grok.java
```diff
@@ -0,0 +1,204 @@
...
+        printLine();
+        System.out.println("Hello, I am Grok! Your favourite personal assistant that helps you keep track of tasks :)");
+        System.out.println("Here are the list of things Grok can do for you:");
+        System.out.println("1. Create a todo task eg. [todo read book]");
+        System.out.println("2. Create an event task eg. [event read book /from 2pm /to 4pm]");
+        System.out.println("3. Create a deadline task eg. [deadline read book /by 2pm]");
+        System.out.println("4. Type either mark or unmark and the task number to indicate completion of task");
+        System.out.println("5. Type delete followed by the task number to remove a task from your list");
+        System.out.println("6. Type list to view your list of tasks.");
+        System.out.println("7. Type bye to exit the programme");
```
okkhoy: Can extract this to a method

src/main/java/grok/Grok.java
```diff
@@ -0,0 +1,204 @@
...
+                    tasks[taskCount] = new Event(taskDetails[2], taskDetails[3], taskDetails[4]);
+                }
+
+                if (isDone) {
+                    tasks[taskCount].markAsDone();
+                }
+                taskCount++;
+            }
+            fileScanner.close();
+        } catch (IOException e) {
```
okkhoy, line 49: consider catching multiple exceptions here related to file operations.

src/main/java/grok/Grok.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        scanner.close();
+    }
+
+    public static void printLine() {
+        System.out.println("____________________________________________________________");
+    }
+
+    private static void saveTasksToFile() {
```
okkhoy: all non-trivial methods need header comments

okkhoy (commented):
consider refactoring main run loop

PR #89, opened 2024-09-02
[shyaamald] iP

shyaamald:


src/main/java/Transcendent.java
```diff
@@ -0,0 +1,178 @@
...
+
+        private static void printEntry() {
+            Printer.printSeparator();
+            System.out.println("Greetings.");
+            System.out.println("I am Transcendent.");
+            System.out.println("What do you need assistance with?");
+            Printer.printSeparator();
+        }
+
+        private static void printSeparator() {
```
gvnchua: good job making the printSeparator() as a separate method to cut down on repeating it

src/main/java/Transcendent.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        private static void printExit() {
+            Printer.printSeparator();
+            System.out.println("Let me know should you need assistance again.");
+            System.out.println("Farewell.");
+            Printer.printSeparator();
+        }
+
+        private static void printAddConfirm(String task) {
```
gvnchua: is there any reason for adding the word confirm for adding and removing tasks? Could the naming be a bit more consistent overall

src/main/java/Transcendent.java
```diff
@@ -0,0 +1,178 @@
...
+                return (isDone ? "[X]" : "[ ]");
+            }
+        }
+
+        private static int listCount = 0;
+
+        private static Task[] tasks;
+
+        private static void init() {
+            tasks = new Task[100];
```
gvnchua: could the number 100 be clearer? why is the array size 100?

src/main/java/Transcendent.java
```diff
@@ -0,0 +1,178 @@
...
+                    break;
+                }
+                if (command.equals("list")) {
+                    List.list();
+                } else if (command.startsWith("mark ")) {
+                    String numberString = command.split(" ")[1];
+                    int indexToBeMarked;
+                    try {
+                        indexToBeMarked = Integer.parseInt(numberString) - 1;
+                        if (indexToBeMarked >= List.listCount) {
```
gvnchua: could the code here be cleaner so as to avoid too many nested statements?

src/main/java/Transcendent.java
```diff
@@ -0,0 +1,178 @@
...
+                Printer.printAlreadyUnmarked();
+            } else {
+                tasks[toBeUnmarked].isDone = false;
+                Printer.printUnmarkConfirm(toBeUnmarked);
+            }
+        }
+
+    }
+
+}
```
gvnchua: overall, good job, maybe the classes could be separated into different files so it is more readable but well done overall

src/main/java/Transcendent.java
```diff
@@ -0,0 +1,293 @@
...
+        private static void printInvalidTodo() {
+            Printer.printSeparator();
+            System.out.println("Invalid todo.");
+            System.out.println("Todo creation syntax: ");
+            System.out.println("todo *task*");
+            System.out.println("For list of all valid instructions: help");
+            Printer.printSeparator();
+        }
+
+        private static void printInvalidEvent() {
```
okkhoy: all non-trivial methods need header comments

src/main/java/Transcendent.java
```diff
@@ -0,0 +1,293 @@
...
+            System.out.println("To add a event: event *task* /from *from* /to *to*");
+            System.out.println("To mark a task: mark *task_number*");
+            System.out.println("To unmark a task: unmark *task_number*");
+            System.out.println("To list all tasks: list");
+            System.out.println("To exit: bye");
+            Printer.printSeparator();
+        }
+    }
+
+    private static class InputHandler {
```
okkhoy: Create individual classes rather than nested classes.

PR #95, opened 2024-09-03
[DarkDragoon2002] iP

DarkDragoon2002:


src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
+import java.util.Scanner;
+
+public class Juan {
+    public static void main(String[] args) {
+        lineMessage();
+        helloMessage();
+        lineMessage();
+
+        boolean continueChatting = true;
```
hzxnancy: Boolean variables/methods should be named to sound like booleans. 
The same error is present in other parts of the code.

src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
...
+        while (continueChatting) {
+            continueChatting = chatFeature();
+        }
+        byeMessage();
+        lineMessage();
+    }
+    public static boolean chatFeature(){
+        // Less efficient to create a new scanner everytime but code is much neater
+        // If return True means continue
+        // Else End
```
hzxnancy: This comment looks like a note to self and should not be here

src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
...
+public class Juan {
+    public static void main(String[] args) {
+        lineMessage();
+        helloMessage();
+        lineMessage();
+
+        boolean continueChatting = true;
+        while (continueChatting) {
+            continueChatting = chatFeature();
+        }
```
hzxnancy: Would it be better to extract these lines of code?

src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
...
+            "       .'     |-_-_-_-|      `.\n" +
+            "      (       `.-_-_-.'        )\n" +
+            "      !`.                    .'!\n" +
+            "        ! ` .            . ' !\n" +
+            "          ! ! ! ! ! ! ! !  !\n" +
+            "            / /       \\ \\\n" +
+            "          _-| \\___ ___/ /-_\n" +
+            "         (_ )__\\_)\\(_/__( _)\n" +
+            "             ))))\\X\\ ((((\n" +
+            "               \\/ \\/ \n" +
```
hzxnancy: I like this logo, it is very cute!

hzxnancy (commented):
Overall, the code looks neat with the extraction of different functions and the use of constructors. The personalisation is very unique which makes this code fun to read, gran trabajo! 

However, do take note of the naming of boolean variables.

src/main/java/Juan.java
Sukkaito: ASCII art rules! 👍

src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
...
+            Task.mark(taskIndex);
+        } else if (parts[0].equals("unmark")){
+            // Unmark
+            int taskIndex = Integer.parseInt(parts[1]) - 1;
+            Task.unmark(taskIndex);
+        } else {
+            // else add task
+            Task newTask = new Task(line);
+        }
+
```
Sukkaito: Should this part be extracted out?

src/main/java/Task.java
```diff
@@ -0,0 +1,66 @@
+public class Task {
+
+    // Keep track of tasks
+    private static Task[] tasks = new Task[100];
+    private static int taskNumber = 0;
```
Sukkaito: Perhaps a better name can be made here, like `taskCount`?

src/main/java/Task.java
```diff
@@ -0,0 +1,66 @@
...
+
+    // Object specific variables
+    public String taskString;
+    public boolean taskDone = false;
```
Sukkaito: Should the boolean name here be more intuitive like in the Java coding standard?

Sukkaito (commented):
In general, LGTM! 👍

src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
...
+        lineMessage();
+        return true;
+    }
+
+    // Message Functions for cleaner main Function
+    public static void lineMessage() {
+        String line = "____________________________________________________________\n";
+        System.out.print(line);
+    }
```
DanLinhHuynh-Niwashi: For the messages, should it be static variables instead and what the methods do is just printing it out?

src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
...
+    public static boolean chatFeature(){
+        // Less efficient to create a new scanner everytime but code is much neater
+        // If return True means continue
+        // Else End
+
+        Scanner scanner = new Scanner(System.in);
+        String line = scanner.nextLine();
+        lineMessage();
+
+        if (line.equals("bye")) {
```
DanLinhHuynh-Niwashi: May the command words also be variables, for later use or easier management?

src/main/java/Juan.java
DanLinhHuynh-Niwashi: Love the logo!

src/main/java/Juan.java
```diff
@@ -0,0 +1,80 @@
...
+        lineMessage();
+
+        boolean continueChatting = true;
+        while (continueChatting) {
+            continueChatting = chatFeature();
+        }
+        byeMessage();
+        lineMessage();
+    }
+    public static boolean chatFeature(){
```
DanLinhHuynh-Niwashi: As it's a method, should it be named with a verb?

src/main/java/Juan.java
```diff
@@ -0,0 +1,139 @@
...
+        lineMessage();
+        helloMessage();
+        lineMessage();
+
+        // Continue chatting as long as user doesn't exit
+        boolean continueChatting = true;
+        while (continueChatting) {
+            // Chat feature to handle user input
+            continueChatting = chatFeature();
+        }
```
wenxin-c: You can consider extracting this part further to match same level of abstraction in this method.
E.g.
```
readData();
salary = basic * rise + 1000;
tax = (taxable ? salary * 0.07 : 0);
displayResult();
```
versus
```
readData();
processData();
displayResult();
```

src/main/java/Juan.java
```diff
@@ -0,0 +1,139 @@
...
+        // Handle "bye" command to end chat
+        if (line.equals("bye")) {
+            return false; // End the conversation
+        }
+        // Handle "list" command to display the list of tasks
+        else if (line.equals("list")) {
+            Task.printTasksList(); // Print task list from Task class
+        }
+        // Handle "mark" command to mark a task as completed
+        else if (line.startsWith("mark ")) {
```
wenxin-c: Try to avoid magic literals(e.g. numbers, strings), you can consider using named constants instead. It will be useful especially when the particular literals are used multiple times in the code.
E.g.
```
static final double PI = 3.14236;
static final int MAX_SIZE = 10;
...
return PI;
...
return MAX_SIZE - 1;

```

src/main/java/Juan.java
```diff
@@ -0,0 +1,139 @@
...
+            continueChatting = chatFeature();
+        }
+
+        // Display goodbye message when the chat ends
+        byeMessage();
+        lineMessage();
+    }
+
+    // Handles user input and executes corresponding actions
+    public static boolean chatFeature() {
```
wenxin-c: This method is very long, you can consider extracting the details for each case into separate methods. It will be better to keep each method short.

src/main/java/Juan.java
```diff
@@ -0,0 +1,139 @@
...
+                // Handle invalid index or format exceptions
+                System.out.println(porFavor + "MARK EXCEPTION: INVALID TASK INDEX");
+            } catch (NullPointerException e) {
+                // Handle null task case
+                System.out.println(porFavor + "MARK EXCEPTION: NULL TASK INDEX");
+            }
+        }
+        // Handle "unmark" command to unmark a task as completed
+        else if (line.startsWith("unmark ")) {
+            try {
```
wenxin-c: Try to keep else if and closing curly bracket on the same line.
E.g.
```
if (condition) {
    statements;
} else if (condition) {
    statements;
} else {
    statements;
}

```

wenxin-c (commented):
Overall good job so far!! Just a few suggestions for you to consider.

src/main/java/Juan.java
DarkDragoon2002: Thanks for the feedback just updated this!

src/main/java/Juan.java
DarkDragoon2002: Thanks for the feedback just updated this! I extracted it into another class to make it easier to read

src/main/java/Juan.java
DarkDragoon2002: Thanks for the feedback just updated this!

PR #105, opened 2024-09-03
[Shen Jiaming] iP

Oshen27:


src/main/java/Bron.java
```diff
@@ -0,0 +1,54 @@
+import java.util.Scanner;
+
+public class Bron {
+    public static void main(String[] args) {
+        String logo = "  ____    ____      ____    _     _\n"
+                    + " |  _ \\  |   _\\   /  __  \\ | \\   | |\n"
+                    + " | |_) | | |_) |  | |  | | |  \\  | | \n"
+                    + " |  _ <  | ___/   | |  | | | |\\\\ | |\n"
+                    + " | |_) | | | \\ \\  | |__| | | | \\\\| |\n"
+                    + " |____/  |_|  \\_\\  \\____/  |_|  \\__|\n";
```
DanLinhHuynh-Niwashi: Should this be a static variable?

src/main/java/Bron.java
```diff
@@ -0,0 +1,54 @@
...
+        Scanner input = new Scanner(System.in);
+        System.out.println(logo + "Hello! I'm Bron\n" + "What can I do for you?\n");
```
DanLinhHuynh-Niwashi: You may change your name into a variable for easier reuse & modification (if any)

src/main/java/Bron.java
```diff
@@ -0,0 +1,54 @@
...
+                    tasks[taskIndex].markAsNotDone();
+                    System.out.println("Get yo shit together son, this task aint done yet:");
+                    System.out.println("  " + tasks[taskIndex].printTask());
+                } else {
+                    System.out.println("Task not found.");
+                }
+            } else {
+                System.out.println("added" + ": " + line);
+                tasks[taskCount++] = new Task(line);
+            }
```
DanLinhHuynh-Niwashi: You may separate the process of each command into methods for readability and reusability.

Should the command words (ex: "mark", "list", ...) be static variables?

src/main/java/Task.java
```diff
@@ -0,0 +1,25 @@
...
+        this.isDone = true;
+    }
+
+    public void markAsNotDone() {
+        this.isDone = false;
+    }
+
+    public String printTask() {
+        return "[" + getStatusIcon() + "] " + this.description;
+    }
```
DanLinhHuynh-Niwashi: As the method is returning a String, not printing it out, should its name be changed to getTaskString() or something like that?

src/main/java/Bron.java
Hackin7: Should `logo` be capitalised as `LOGO` instead to follow the naming convention for constants?

src/main/java/Bron.java
```diff
@@ -0,0 +1,54 @@
...
+                    + " |  _ <  | ___/   | |  | | | |\\\\ | |\n"
+                    + " | |_) | | | \\ \\  | |__| | | | \\\\| |\n"
+                    + " |____/  |_|  \\_\\  \\____/  |_|  \\__|\n";
+        Scanner input = new Scanner(System.in);
+        System.out.println(logo + "Hello! I'm Bron\n" + "What can I do for you?\n");
+
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+
+        while (true) {
```
Hackin7: Should the main function be shortened (eg. with more abstraction)? This `main` function is quite hard to read for me as it is quite long.

text-ui-test/runtest.sh
Hackin7, line 1: Should there be something in this file to prevent people from being mislead, or at least, a TODO in this file to understand what needs to be done?

src/main/java/Bron.java
```diff
@@ -0,0 +1,54 @@
...
+                System.out.println("Catch you on the flip cuh");
+                break;
+            }
+
+            if (line.equalsIgnoreCase("list")) {
+                int listCount = 0;
+                while(tasks[listCount] != null) {
+                    System.out.println(listCount + 1 + ". " + tasks[listCount++].printTask());
+                }
+            } else if (line.startsWith("mark")){
```
Hackin7: This looks like arrowhead code, and I feel with more abstraction we could reduce this effect

src/main/java/bron/Bron.java
```diff
@@ -0,0 +1,33 @@
...
+
+        Task[] tasks = initializeTasks();
+        int taskCount = 0;
+
+        ChatBotManager chatBotManager = new ChatBotManager(tasks, taskCount);
+        chatBotManager.start();
+    }
+
+    private static Task[] initializeTasks() {
+        return new Task[100];
```
dZhei98: Avoid the use of magic numbers.

PR #106, opened 2024-09-03
[Alfred Goh] iP

Alfred-Goh02:


src/main/java/Poppy.java
```diff
@@ -0,0 +1,155 @@
...
+
+        public String taskType(){
+            return "[]";
+        }
+        //...
+    }
+    public static class ToDo extends Task{
+        public ToDo (String description){
+            super(description);
+            System.out.println("[T]" + " [" + super.getStatusIcon()+ "] " + description);
```
C2linaung: I think it would be better to standardise the spacings.. especially between the plus sign

src/main/java/Poppy.java
```diff
@@ -0,0 +1,155 @@
...
+        public Task(String description) {
+            this.description = description;
+            this.isDone = false;
+        }
+
+        public String getStatusIcon() {
+            return (isDone ? "X" : " "); // mark done task with X
+        }
+
+        public void markAsDone() {
```
C2linaung: using "setDone" is better for function naming. for IP i think is ok to do this since its individual work anyways, but when we doing tP, it would be good to take note for this since we want the code to look like its written by one person.

src/main/java/Poppy.java
```diff
@@ -0,0 +1,155 @@
...
+            System.out.println("[T]" + " [" + super.getStatusIcon()+ "] " + description);
+        }
+
+        @Override
+        public String taskType() {
+            return "[T] ";
+        }
+    }
+
+    public static class Deadline extends Task{
```
C2linaung: Will it be better to put the classes into different files?

src/main/java/Poppy.java
```diff
@@ -0,0 +1,155 @@
...
+        System.out.println("Hello! I'm Poppy");
+        System.out.println("What can I do for you?");
+        Task[] list = new Task[100];
+        input = sc.nextLine();
+        list[0]= new Task(input);
+        int counter = 0;
+        while (!input.equals("Bye")) {
+            String[] str = input.split(" ", 2 );
+            String command = str[0];
+            switch (command) {
```
C2linaung: I like the use of switch cases here instead of if cases since there are many different conditions.

src/main/java/Poppy.java
```diff
@@ -0,0 +1,155 @@
...
+        public String taskType() {
+            return "[D] ";
+        }
+    }
+
+    public static class Events extends Task{
+        protected String from;
+        public Events (String description, String from){
+            super(description);
+            this.from =from;
```
C2linaung: do take note of the spacings here

C2linaung (commented):
Generally good job on the codes. I think some suggestions would be to move the codes into different files to  make it easier to organise stuff and also do take note of the spacings

src/main/java/commands/Commands.java
```diff
@@ -0,0 +1,51 @@
+package commands;
+
+import tasks.*;
```
okkhoy: don't import using wild card importing.

src/main/java/commands/Commands.java
```diff
@@ -0,0 +1,51 @@
...
+                System.out.println((i+1) + ". " + list.get(i).toString());
+            }
+        }
+    }
+
+    public static void markAsNotDone(ArrayList<Task> list, String[] str) {
+        int index = Integer.parseInt(str[1])- 1;
+        if (list.get(index) == null){
+            System.out.println("Sorry, list number does not exist");
+        }else {
```
okkhoy: spacing issue

src/main/java/poppy/Printer.java
```diff
@@ -0,0 +1,171 @@
...
+                case "mark":
+                    markAsDone(taskList, commandArgs);
+                    break;
+                case "unmark":
+                    markAsNotDone(taskList, commandArgs);
+                    break;
+                case "List":
+                    showList(taskList);
+                    break;
+                case "todo":
```
okkhoy: `case` statements are not following our coding standards

src/main/java/poppy/Printer.java
```diff
@@ -0,0 +1,171 @@
...
+        }
+    }
+
+    public static void processFile(String input, ArrayList<Task> taskList) {
+        String[] commandArgs = input.split(" ", 3);
+        if (commandArgs.length < 3) return; // Ensure there are enough arguments
+
+        String command = commandArgs[0];
+        String status = commandArgs[1];
+        String description = commandArgs[2];
```
okkhoy: can improve SLAP here

src/main/java/poppy/Printer.java
```diff
@@ -0,0 +1,171 @@
...
+        while (!input.equals("Bye")) {
+            input = userInput.nextLine();
+            processCommands(input, taskList);
+            writeToFile(file, taskList);
+        }
+        System.out.println("Bye. Hope to see you again soon!");
+        userInput.close();
+    }
+
+    private static void writeToFile(File file, ArrayList<Task> taskList) throws IOException {
```
okkhoy: All non-trivial methods need header comments

okkhoy (commented):
I noticed some coding standard violations. Please fix them before the final submission

PR #108, opened 2024-09-04
[andreusxcarvalho] iP

andreusxcarvalho:


src/main/java/Andy.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private static void greetUser() {
+        System.out.println("_______________________________________");
+        System.out.println("Hello! I'm ANDY");
+        System.out.println("What can I do for you?");
+        System.out.println("Type 'list' followed by an item to add it to the list.");
+        System.out.println("Type 'list show' to display all items in the list.");
+        System.out.println("Type 'mark <number>' to mark a task as done.");
+        System.out.println("Type 'unmark <number>' to mark a task as not done.");
+        System.out.println("_______________________________________");
```
Hackin7: Would it be better to store the greeting string as a constant instead of directly printing it?
In the event we need to change the `System.out.println` function to something else it would be easier + avoid magic number scenario

src/main/java/Andy.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    private static void exit() {
+        System.out.println("_______________________________________");
+        System.out.println("Bye. Hope to see you again soon!");
+        System.out.println("_______________________________________");
+    }
+
+    // Task class to represent each task in the list
+    static class Task {
```
Hackin7: Should multiple classes be separated into different files for readability?

src/main/java/Andy.java
```diff
@@ -0,0 +1,136 @@
...
+            System.out.println("Please provide a valid list command or item.");
+        }
+    }
+
+    private static void addItemToList(String item) {
+        itemList.add(new Task(item));
+        System.out.println("Item added to the list: " + item);
+    }
+
+    private static void showList() {
```
Hackin7: The abstraction of the various features into various functions made the code easier to read. Well done!

src/main/java/Andy.java
```diff
@@ -0,0 +1,136 @@
...
+        System.out.println("Type 'list show' to display all items in the list.");
+        System.out.println("Type 'mark <number>' to mark a task as done.");
+        System.out.println("Type 'unmark <number>' to mark a task as not done.");
+        System.out.println("_______________________________________");
+    }
+
+    private static void handleListInput(String input) {
+        String[] parts = input.split(" ", 2); // Split the input into command and item
+
+        if (parts.length > 1) {
```
Hackin7: I think you can avoid arrowhead style code by using guard clauses. In each if statement, you return after the function was run. This way, you don't need an if else if chain

src/main/java/Andy.java
```diff
@@ -0,0 +1,136 @@
...
+                markTaskAsDone(input);
+            } else if (input.startsWith("unmark ")) {
+                markTaskAsNotDone(input);
+            } else {
+                echo(input); // Call echo method to repeat the input
+            }
+        }
+
+        exit();
+    }
```
gskang-22: I like how your statements are at the same level of abstraction

src/main/java/Andy.java
gskang-22: Would it be better to use a single println() function with concatenation instead of having multiple println() functions?

src/main/java/Andy.java
```diff
@@ -0,0 +1,136 @@
...
+        public void setDone(boolean done) {
+            isDone = done;
+        }
+
+        @Override
+        public String toString() {
+            return (isDone ? "[X] " : "[ ] ") + description;
+        }
+    }
+}
```
gskang-22: Would it be better to move your classes into their own files (AKA Task.java) to make your code more organized?

src/main/java/Andy.java
```diff
@@ -0,0 +1,136 @@
...
+    private static void markTaskAsDone(String input) {
+        try {
+            int taskNumber = Integer.parseInt(input.split(" ")[1]) - 1;
+            itemList.get(taskNumber).setDone(true);
+            System.out.println("Nice! I've marked this task as done:");
+            System.out.println("[X] " + itemList.get(taskNumber).getDescription());
+        } catch (Exception e) {
+            System.out.println("Invalid task number. Please try again.");
+        }
+    }
```
gskang-22: I like how you use try and catch statements to catch errors and prevent your code from crashing

gskang-22 (commented):
Good work. Overall your code looks concise and free of mistakes.

src/main/java/Andy.java
```diff
@@ -0,0 +1,173 @@
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Scanner;
+
+public class Andy {
+
+    private static List<Task> itemList = new ArrayList<>(); // List to store tasks
```
dZhei98: This can be renamed to taskList to be more descriptive.

PR #109, opened 2024-09-04
[Guoyi] ip

gu0y1:


src/main/java/Duke.java
```diff
@@ -6,5 +8,129 @@ public static void main(String[] args) {
...
+        System.out.println("What can I do for you?");
+
+        Scanner scanner = new Scanner(System.in);
+        Task[] tasks = new Task[100];
+        int taskCount = 0;
+        String input;
+
+        while (true) {
+            input = scanner.nextLine();
+            if (input.equals("bye")) {
```
ryryry-3302, line 21: Instead of using a long list of else ifs you could consider re factoring this into switch case form to improve readability. Just a suggestion!

src/main/java/Duke.java
```diff
@@ -6,5 +8,129 @@ public static void main(String[] args) {
...
+            }
+        }
+
+        scanner.close();
+    }
+}
+
+class Task {
+    protected String description;
+    protected boolean isDone;
```
ryryry-3302, line 71: Great job following the naming conventions for boolean variables!

src/main/java/Duke.java
```diff
@@ -6,5 +8,129 @@ public static void main(String[] args) {
...
+            } else if (input.startsWith("todo ")) {
+                String description = input.substring(5);
+                tasks[taskCount] = new ToDo(description);
+                taskCount++;
+                System.out.println("Got it. I've added this task:");
+                System.out.println("  " + tasks[taskCount - 1]);
+                System.out.println("Now you have " + taskCount + " tasks in the list.");
+            } else if (input.startsWith("deadline ")) {
+                String[] parts = input.substring(9).split(" /by ");
+                tasks[taskCount] = new Deadline(parts[0], parts[1]);
```
ryryry-3302, line 48: You could consider adding some edge case catchers here to remind user the correct argument usage. 
For example, if the user forgets to put /by you could print out something like expected argument containing /by

src/main/java/Duke.java
```diff
@@ -6,5 +8,129 @@ public static void main(String[] args) {
...
+    }
+
+    @Override
+    public String toString() {
+        return "[D]" + super.toString() + " (by: " + by + ")";
+    }
+}
+
+class Event extends Task {
+    protected String from;
```
ryryry-3302, line 123: Great job keeping all variables in protected accessibility to improve security of your object's members!

ryryry-3302 (commented):
Looks good just some minor suggestions!

src/main/java/Duke.java
```diff
@@ -6,5 +8,129 @@ public static void main(String[] args) {
                 + "| |_| | |_| |   <  __/\n"
                 + "|____/ \\__,_|_|\\_\\___|\n";
         System.out.println("Hello from\n" + logo);
+        System.out.println("Hello! I'm Duke");
+        System.out.println("What can I do for you?");
+
+        Scanner scanner = new Scanner(System.in);
+        Task[] tasks = new Task[100];
```
joshuan98, line 15: You can consider the case where there are more than 100 tasks. ArrayList may be more applicable

dZhei98, line 15: You can avoid the use of 100 here which is a magic number.

src/main/java/Duke.java
```diff
@@ -6,5 +8,129 @@ public static void main(String[] args) {
...
+                System.out.println("Bye. Hope to see you again soon!");
+                break;
+            } else if (input.equals("list")) {
+                System.out.println("Here are the tasks in your list:");
+                for (int i = 0; i < taskCount; i++) {
+                    System.out.println((i + 1) + "." + tasks[i]);
+                }
+            } else if (input.startsWith("mark ")) {
+                int taskIndex = Integer.parseInt(input.split(" ")[1]) - 1;
+                tasks[taskIndex].markAsDone();
```
joshuan98, line 31: Consider the case where the taskIndex provided is out-of-bounds. Error handling may be necessary.

dZhei98 (commented):
Instead of putting all your code in a single file, consider using separate files instead.

PR #110, opened 2024-09-04
[Zhitong] ip

Zhang-Zhitong:


src/main/java/Commands.java
```diff
@@ -0,0 +1,60 @@
...
+    private boolean isExited = false;
+
+    public Commands() {
+
+    }
+
+    public void accept(String nextLine) {
+
+        int index = (nextLine.indexOf(' '));
+        String commandKeyword = (index == -1)? nextLine : nextLine.substring(0, index).trim();
```
RCPilot1604, line 19: Should there be a space before the "?" ternary operator?

src/main/java/Texts.java
```diff
@@ -0,0 +1,6 @@
+
+public class Texts {
+    public static final String GREETINGS = " hello";
```
RCPilot1604, line 3: I like how you use all caps for constants.

src/main/java/Task.java
```diff
@@ -0,0 +1,36 @@
...
+
+    public Task(String line) {
+        this.line = line;
+        this.isDone = false;
+    }
+
+    @Override
+    public String toString() {
+        return "[" + (isDone? "X" : " ") + "]"
+                + line;
```
RCPilot1604, line 34: Shouldn't this be on the same line (line limit 110 characters)?

src/main/java/Commands.java
```diff
@@ -0,0 +1,60 @@
...
+
+        final String commandText = (index == -1)? nextLine : nextLine.substring(index).trim();
```
RCPilot1604, line 21: Shouldn't you extract this magic number?

src/main/java/Commands.java
```diff
@@ -0,0 +1,60 @@
+
+
+public class Commands {
+
+    private static final int MAX_LENGTH = 100;
+    private Task[] tasks = new Task[MAX_LENGTH];
+
+    public int length = 0;
+
+    private boolean isExited = false;
```
RCPilot1604, line 10: Shouldn't the private attributes be grouped together?

RCPilot1604 (commented):
Nice work! :)

okkhoy (commented):
not much code to review.
do not track binary files in git.

PR #113, opened 2024-09-04
[Ayushi Yadav] IP

ayushi0803:


src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+
 public class Duke {
     public static void main(String[] args) {
-        String logo = " ____        _        \n"
-                + "|  _ \\ _   _| | _____ \n"
-                + "| | | | | | | |/ / _ \\\n"
-                + "| |_| | |_| |   <  __/\n"
-                + "|____/ \\__,_|_|\\_\\___|\n";
-        System.out.println("Hello from\n" + logo);
+        System.out.println("____________________________________________________________");
```
Sukkaito, line 5: I think you can just keep the indentation of these "separator line" (?), or extract them into a separate function and call the function instead.

src/main/java/Task.java
```diff
@@ -0,0 +1,28 @@
+public class Task {
+  protected String description;
+  protected boolean isDone;
+  protected String type;
```
Sukkaito, line 4: Perhaps a better name for `type` could be more intuitive here?

src/main/java/Task.java
Sukkaito, line 1: Should the indentation for this file be 4 spaces instead of 2?

src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+            } else if (input.startsWith("event")) {
+                String taskDescription = input.substring(6);
+                tasks[taskCount] = new Task(taskDescription, "E");
+                taskCount++;
+                System.out.println("added: " + taskDescription);
+                System.out.println("____________________________________________________________");
+            }
+        }
+
+        scanner.close();
```
Sukkaito, line 61: Nice! You did remember to close the `scanner`! 👍 (Not like me)

Sukkaito (commented):
In general, LGTM! 👍

src/main/java/Duke.java
Yvorm: I think this is a neat idea - especially if you want to modularise your responses later down the line.

src/main/java/Duke.java
```diff
@@ -1,10 +1,63 @@
...
+                System.out.println("added: " + taskDescription);
+                System.out.println("____________________________________________________________");
+            } else if (input.startsWith("event")) {
+                String taskDescription = input.substring(6);
+                tasks[taskCount] = new Task(taskDescription, "E");
+                taskCount++;
+                System.out.println("added: " + taskDescription);
+                System.out.println("____________________________________________________________");
+            }
+        }
```
Yvorm, line 59: I like the structure of this segment! I noticed some of the repeated could be abstracted into a function call. Were you planning on adding unique functionality to each case?

